PR #5, opened 2023-05-01
tech: add SearchRestController

p4dd9:


p4dd9:
hydrate frontend search from url query params

PR #60, opened 2023-09-22
(docs): add local development setup steps

cstenglein:
This removes the default README instructions and adds correct local setup steps to it


kadhonn:
thank you so much!

PR #88, opened 2023-10-04
added stiftskonzerte collector

belehla:
issue #44

### DIFF
--- a/eventcollectors/src/main/kotlin/events/boudicca/eventcollector/collectors/StiftskonzerteCollector.kt
+++ b/eventcollectors/src/main/kotlin/events/boudicca/eventcollector/collectors/StiftskonzerteCollector.kt
@@ -1,0 +1,65 @@
+package events.boudicca.eventcollector.collectors
+
+import events.boudicca.SemanticKeys
+import events.boudicca.api.eventcollector.Event
+import events.boudicca.api.eventcollector.Fetcher
+import events.boudicca.api.eventcollector.TwoStepEventCollector
+import org.jsoup.Jsoup
+import org.jsoup.nodes.Element
+import java.time.LocalDate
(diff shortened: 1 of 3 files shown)


eventcollectors/src/main/kotlin/events/boudicca/eventcollector/collectors/StiftskonzerteCollector.kt
```diff
@@ -0,0 +1,65 @@
...
+    override fun parseEvent(event: String): Event {
+        val eventSite = Jsoup.parse(fetcher.fetchUrl(event))
+        val name = eventSite.select("header.entry-header h1").text()
+
+        val data = mutableMapOf<String, String>()
+        data[SemanticKeys.URL] = event
+        data[SemanticKeys.DESCRIPTION] = eventSite.select("div.entry-flexible-content").text()
+
+        val locationTimeDiv = eventSite.select("div.entry-content div.location")
+        locationTimeDiv.select("br").after("\\n")
```
kadhonn, line 34: does this line actually do something?

kadhonn (commented):
thanks for the second collector :)

eventcollectors/src/main/kotlin/events/boudicca/eventcollector/collectors/StiftskonzerteCollector.kt
belehla: yes, it helps to preserve the line breaks by adding \n after each br tag, because the br tags disappear when using .text() later

kadhonn: ohh, awesome! thanks for explaining

PR #97, opened 2023-10-07
fix some minor suggestions and warnings and deprecations

Yolgie:


eventcollectors/build.gradle.kts
```diff
@@ -15,6 +15,7 @@ dependencies {
         exclude("org.codehaus.groovy", "groovy-dateutil")
     }
     implementation("com.rometools:rome:2.1.0")
+    implementation("org.apache.httpcomponents:httpclient:4.5.13")
```
kadhonn, line 18: why did you add this dependency explicitly?

eventcollectors/src/main/kotlin/events/boudicca/eventcollector/collectors/TechnologiePlauscherlCollector.kt
```diff
@@ -49,4 +51,14 @@ class TechnologiePlauscherlCollector : EventCollector {
 
         return events
     }
+
+    private fun prepareSyndFeed(url: String): SyndFeed {
+        HttpClients.createMinimal().use { client ->
```
kadhonn, line 56: please use our Fetcher class for all http relevant stuff, as it implements a delay handling, sets the user-agent to boudicca and records http calls automatically in our collectors.boudicca.event overview

kadhonn (changes requested):

eventcollectors/build.gradle.kts
kadhonn: because you used it below...

eventcollectors/src/main/kotlin/events/boudicca/eventcollector/collectors/TechnologiePlauscherlCollector.kt
Yolgie: the XmlReader with an url as param has been deprecated for quite some time, if you can create a synd feed with the Fetcher, pls tell me how to do it, otherwise this would require major refactoring of either the Fetcher or the Techplauscherl Collector

Yolgie: i just converted the fetched content to an InputStream, not the most elegant solution, but it works...

kadhonn (approved):
looks good, thanks for the cleanup!

PR #105, opened 2023-10-08
Abl/split boudicca

### LINKED ISSUE #103: split base boudicca software and boudicca.events software

kadhonn:
ook, so the base idea with this is that we want to separate "the base boudicca code" with our own deployment. i did this by moving relevant pieces into the boudicca.events subfolder.

these pieces are for now:
1) all the eventcollectors we have. not the api for them, but the real collectors. if someone wants to host them he really is welcomed to, but why not just fetch the events from our boudicca installation, would be much easier. and they still can of course easily create their own deployment with their own eventcollectors.
2) i split the publisher-html into a base version and one which only contains an extension, which basically is the three header items

i kinda like the solution i arrived at, but since this is a bigger rework i want your opinion as well @Yolgie @p4dd9 @twatzl 

resolves #103


Yolgie:
i really don't like the repetition in the namespace: `boudicca.boudicca.events. ...`

Yolgie:
ok, i have to look at this in a little more detail than i can right now and check out the branch locally since github does not really show me nicely what was moved...

kadhonn:
> i really don't like the repetition in the namespace: `boudicca.boudicca.events. ...`

yeah but this is only in the gradle projects, not in the actual kotlin-packages or something like that

PR #108, opened 2023-10-09
add bare enricher service

kadhonn:
i added a bare (aka not doing anything right now) enricher server, and the eventcollectors can be configured to call it. does this look like a good thing or would you have implemented it differently?
@Yolgie @p4dd9 @twatzl


Yolgie:
There are multiple ways we can go with this (and they are not even exclusive to each other) 

1. Enricher called by the collector before pushing data (this would be especially interresting for enrichment that is dependent on the source -> music venues doing artist enrichment, maybe doing the CrowdSourced enrichment via a service here)
2. Enricher goes through the DB and enriches every entry that it can (this would make sense for geodata stuff for example)
3. Enricher that is called during the search (this would be interresting when working with searcher information like position -> add public transport options from my location to the event)

So, for your solution (1) I would 

- add the option to pass data to the enricher in the EnricherRequestDTO about the collector itself
- add the option to add multiple enrichers to a collector
- am i reading this wrong or is the scheduler calling the enrichor(s), should the collector not call them?
  - i think i get it, your enricher would be called automatically for all collectors right? so it's actually more the case (2) but at the time of writing into the DB, hmm, has it's advantages, i have to think about this...

Yolgie:
on a side note: java.Funciton & java.Consumer are not really the Kotlin way of doing this -> use lambda descriptions instead :)
https://kotlinlang.org/docs/lambdas.html

Yolgie:
another sidenote: if we are doing real reviews here, we should really get into the habit of writing tests as well 😆

kadhonn:
> another sidenote: if we are doing real reviews here, we should really get into the habit of writing tests as well 😆 

hey, i did at a "does it start" test here :P

Yolgie:
Can you update your example with the MusicBrainzEnricher added? Right now no enricher is configured right?

enricher/src/main/kotlin/events/boudicca/enricher/model/Event.kt
```diff
@@ -0,0 +1,9 @@
+package events.boudicca.enricher.model
+
+import java.time.ZonedDateTime
+
+data class Event(
```
Yolgie, line 5: we really should not duplicate this all over

kadhonn: we should not, but with the current openapi setup we have to do it, because openapi is, once again, a pain...

kadhonn:
> Can you update your example with the MusicBrainzEnricher added? Right now no enricher is configured right?

not really right now because it does not work nearly as good as i hade hoped it would :(
need more time with it
but something with the locations should be easy to add soonish

Yolgie:
the musicBrainzEnricher is a NOP enricher anyways, but currently no enricher is added anywhere

enricher/src/main/kotlin/events/boudicca/enricher/model/Event.kt
Yolgie: yes, we should really take care of the openAPI stuff once and for all...

kadhonn:
and to your comment about the different kinds of enricher...

for now the scheduler calls the enricher for all collectors, because i guess that will be the default?
but i did want to have the decision if data should be enriched at the collectors side, that is why i did not made the enricher just crawl through our database. as i have seen it till now (maybe this is wrong), data should stay as the collector sends it. so if the collector does not want any enrichment we should not do it later on the eventdb side

kadhonn:
sure, the NOP musicbrainz is running for each event submitted. the EnricherService gets a list of all enrichers autowired, so you only have to add a new class implementing the Enricher interface and it will be automatically used

Issue #138, opened 2023-10-18
Enhancing accessibility


mahdikhashan:
When it comes to filtering events based on their accessibility for people, I would like to add icons to the event cards.

PR #144, opened 2023-10-19
Nano/enhancing accessibility

mahdikhashan:
This pr will address https://github.com/boudicca-events/boudicca.events/issues/138.


publisher-html/src/main/resources/templates/events/events.hbs
```diff
@@ -80,6 +80,17 @@
         gap: 4px;
     }
 
+    .accessibility-details {
+        display: flex;
+        background-color: white;
+        color: inherit;
+        position: absolute;
+        left: 8px;
```
p4dd9, line 88: [nit] This project uses CSS variables to organize FE values as consistent as possible. Please have a look at `theme.hbs`, it includes a `--space-medium: 8px` variable that should be re-used for these props

publisher-html/src/main/kotlin/events/boudicca/publisherhtml/PublisherHtmlApplication.kt
```diff
@@ -1,5 +1,6 @@
 package events.boudicca.publisherhtml
 
+import com.github.jknack.handlebars.helper.ConditionalHelpers
```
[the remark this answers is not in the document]
p4dd9: Visual sample: 
image

publisher-html/src/main/resources/templates/events/events.hbs
```diff
@@ -80,6 +80,17 @@
         gap: 4px;
     }
 
+    .accessibility-details {
+        display: flex;
+        background-color: white;
+        color: inherit;
```
p4dd9, line 86: [nit] Is there a need to inherit the color in this case? (the stroke color seems to default to black)

publisher-html/src/main/resources/templates/events/event.hbs
```diff
@@ -1,39 +1,50 @@
...
+            <a class="anchor-to-eventpage" href="{{event.url}}" target="_blank"
+               aria-describedby="Zur Eventseite von {{event.name}} (neues Fenster wird geöffnet)">
+                Zur Eventseite
+            </a>
+        {{/if}}
+
+        {{#or event.accessibleEntry event.accessibleSeats event.accessibleToilets }}
+            <div class="accessibility-details">
+                <svg height="24px" width="24px" xmlns="http://www.w3.org/2000/svg" id="Layer_1" data-name="Layer 1" viewBox="0 0 122.88 122.88"><title>
+                    accessibility</title>
```
p4dd9, line 20: [nit] https://developer.mozilla.org/en-US/docs/Web/SVG/Element/title should describe objects by its visual appearance, in this case I would suggest to add a descriptive way of what the vector graphic shows e.g. "schwarze Silhouette einer Person"

publisher-html/src/main/kotlin/events/boudicca/publisherhtml/PublisherHtmlApplication.kt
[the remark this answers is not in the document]
p4dd9: I like the visual prominence of the accessibility indicator, nevertheless let's be careful to not overrule the intend to an interactive button "Zur Eventseite" :-) It already makes me add an overlay to display accessibility features, which is (IMO) the most important information here

kadhonn:
since we cannot update your PR because we do not have rights in your repository, we took the branch and rebased it onto main, see PR #173

PR #171, opened 2023-10-28
update architecture section and add architecture diagram

twatzl:
look at it.. it's fancy..

such wow. much awesome.


kadhonn (approved):
nice, thank you!

PR #180, opened 2023-10-31
extract apis as interfaces

twatzl:


build.gradle.kts
```diff
@@ -1,5 +1,12 @@
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
+
 plugins {
+    id("io.cloudflight.autoconfigure-gradle") version "1.0.0"
```
kadhonn, line 4: i am not quite sold on using these plugins, what exactly are the doing and why do we need them?

kadhonn:
seems to work when i try it locally

PR #238, opened 2023-12-26
fix generation of swaggerdoc for search-api

twatzl:


boudicca.base/search-api/build.gradle.kts
```diff
@@ -2,10 +2,12 @@ plugins {
     kotlin("jvm")
     kotlin("kapt")
     id("io.cloudflight.autoconfigure.swagger-api-configure")
+    `maven-publish`
```
kadhonn, line 5: those lines have nothing to do with the fix, do they? please remove them from the pr, otherwise it looks good, you can just merge them

PR #251, opened 2024-01-14
add generation of npm clients to openapi modules

twatzl:


boudicca.base/enricher-openapi/build.gradle.kts
```diff
@@ -35,15 +33,40 @@ dependencies {
...
-tasks.withType<org.openapitools.generator.gradle.plugin.tasks.GenerateTask> {
+tasks.withType<GenerateTask> {
+    inputs.files(openapi)
+    inputSpec.set(openapi.files.first().path)
+}
+
+tasks.register<GenerateTask>("generateJavaClient") {
     inputs.files(openapi)
     inputSpec.set(openapi.files.first().path)
+    outputDir.set(layout.buildDirectory.dir("/generated/java").get().toString())
```
kadhonn, line 44: i think this is interpreted as an absolute path, so remove the first / or try ./generated

Yolgie:
why did you add the generated code as well?

Yolgie (changes requested):
looks roughly good, but i would not commit the generated code

- [ ] please remove the generated folders from the MR
- [ ] please add the generated folders to the .gitignore

kadhonn:
oook, i kinda got carrried away but i really wanted to do a big cleanup of our gradle stuff. to avoid having merge conflicts with tobis changes i just based mine on his, and also did the cleanup of the typescript files... so that is now a big monster of PR with way too many changes, but i really wanted to get this done. you can have a look at the commits, i tried my best to group them nicely.

anyway, what i did was create "convention-plugins" for our project, where we standardize all the build config for different projects.
so now we have one for springboot apps, openapi client generators, openapi spec generators and some base files as well
that is soo much cleaner already, but there are still some things to be done :)

kadhonn:
i made a mistake, sorry. new MR is #258

PR #263, opened 2024-01-27
replace old spec generator with swagger-core, sadly this means we hav…

kadhonn:
…e to use jax-rs to annotate the api interfaces

through another day of pain i finally have a real and working setup

i kicked out the cloudflight plugin because that was based on an old plugin which was not maintained anymore and only generated swagger 2.0 specs, which our generator could not accept

now we use swagger-core, but that does not have support for spring annotations, so we have to use swagger-annotations and jax-rs to do that for the api and for the spring implementation we use the normal spring annotations... but i could not see any other way to do that.
also contains some hacks because the whole generator is fucking wild (in a bad way)

aaanyway, documentation how the fuck this all works will follow, as soon as i figure out how to publish the openapi files as well on to maven central


twatzl (approved):
Approved!

PR #267, opened 2024-01-27
Abl/docu all the things

kadhonn:


kadhonn:
i added a lost of docu about our gradle plugins and how to use them
and about our rest api, and how to use it. for this especially i want feedback please

you can have a look at https://github.com/boudicca-events/boudicca.events/blob/abl/docu-all-the-things/README.md
and especially https://github.com/boudicca-events/boudicca.events/blob/abl/docu-all-the-things/REST.md

please also have a look at the two sample repos i created which are mentioned in the REST.md docu
https://github.com/boudicca-events/client-samples
https://github.com/boudicca-events/openapi-generate-sample

i know we should add more comments to the sample repos but i am really tired now and i wanted to have some feedback if that is a good direction i am going before putting more effort into it

PR #269, opened 2024-01-31
Add a new collector for valug.at

dmorawetz:
VorAlpen Linux User Group is a local Linux user community that meets regularly. Most events happen at the Klubraum. If the ical event location contains this keyword, further geolocation is added.

The Klubraum is also marked on OpenStreetMap with the current ID of [99255999](https://www.openstreetmap.org/way/99255999). If I understand correctly, [there does not exist any permanent ID](https://wiki.openstreetmap.org/wiki/Permanent_ID) for POIs, except for the wikidata page. The Klubraum or VALUG has no wikidata page. Do you already have a standard how to handle this?

The Klubraum is located on the second floor and only accessible by stairs. Does the absence of `accessibility.accessibleEntry` imply a value of `false` or should this be explicitly set?

### DIFF
--- a/boudicca.events/eventcollectors/src/main/kotlin/events/boudicca/eventcollector/collectors/ValugCollector.kt
+++ b/boudicca.events/eventcollectors/src/main/kotlin/events/boudicca/eventcollector/collectors/ValugCollector.kt
@@ -51,12 +51,6 @@ class ValugCollector : TwoStepEventCollector<VEvent>("valug") {
                 put(SemanticKeys.ENDDATE, eventEndDate.format(DateTimeFormatter.ISO_DATE_TIME))
                 if (event.location != null) {
                     put(SemanticKeys.LOCATION_NAME, event.location.value)
-                    if (event.location.value.contains("VALUG-Klubraum")) {
-                        put(SemanticKeys.LOCATION_CITY, "Wels")
-                        put(SemanticKeys.LOCATION_URL, "https://valug.at/pages/Klubraum/")
-                        put(SemanticKeys.LOCATION_ADDRESS, "Dragonerstraße 22, 4600 Wels")
-                        put(SemanticKeys.LOCATION_COORDINATES, "14.01715, 48.15849")
-                    }
                 }
                 if (event.description != null) {
                     put(SemanticKeys.DESCRIPTION, event.description.value)
--- a/boudicca.events/eventcollectors/src/main/kotlin/events/boudicca/eventcollector/collectors/ValugCollector.kt
+++ b/boudicca.events/eventcollectors/src/main/kotlin/events/boudicca/eventcollector/collectors/ValugCollector.kt
@@ -1,0 +1,75 @@
+package events.boudicca.eventcollector.collectors
+
+import base.boudicca.SemanticKeys
+import base.boudicca.api.eventcollector.Fetcher
+import base.boudicca.api.eventcollector.TwoStepEventCollector
+import base.boudicca.model.Event
+import base.boudicca.model.EventCategory
+import net.fortuna.ical4j.data.CalendarBuilder
+import net.fortuna.ical4j.model.Calendar
+import net.fortuna.ical4j.model.component.VEvent
+import java.io.StringReader
+import java.net.URI
+import java.time.LocalDateTime
+import java.time.format.DateTimeFormatter
+import java.util.*
+
+/**
+ * VorAlpen Linux User Group
+ */
+class ValugCollector : TwoStepEventCollector<VEvent>("valug") {
+    private val fetcher = Fetcher()
+    private val baseUrl = "https://valug.at/"
+    private val icsUrl = "${baseUrl}events/index.ics"
+
+    override fun getAllUnparsedEvents(): List<VEvent> {
+        val feed = fetcher.fetchUrl(icsUrl)
+        val builder = CalendarBuilder()
+        val calendar: Calendar = builder.build(StringReader(feed))
+        return calendar.components.filterIsInstance<VEvent>()
+    }
+
+    override fun parseEvent(event: VEvent): Event? {
+        val eventName = event.summary.value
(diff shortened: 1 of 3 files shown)


boudicca.events/eventcollectors/src/main/kotlin/events/boudicca/eventcollector/collectors/ValugCollector.kt
```diff
@@ -0,0 +1,75 @@
...
+        val eventEndDate = LocalDateTime.parse(event.startDate.value, formatter)
+            .atZone(TimeZone.getTimeZone("Europe/Vienna").toZoneId()).toOffsetDateTime()
+
+        return Event(
+            eventName, eventStartDate,
+            buildMap {
+                put(SemanticKeys.ENDDATE, eventEndDate.format(DateTimeFormatter.ISO_DATE_TIME))
+                if (event.location != null) {
+                    put(SemanticKeys.LOCATION_NAME, event.location.value)
+                    if (event.location.value.contains("VALUG-Klubraum")) {
```
kadhonn, line 54: adding additional location data should be done with our enricher, so that potentially other eventcollectors benefit from that information as well. how that is currently done is by entering this information into this google-sheet, if then the location name matches it will automatically add all the other information there as well: https://docs.google.com/spreadsheets/d/1yYOE5gRR6gjNBim7hwEe3__fXoRAMtREkYbs-lsn7uM/edit#gid=0
i have entered this information for you right now since we do not want to allow anyone to edit this data without moderation.
this means that you please should remove this if from the collector

boudicca.events/eventcollectors/src/main/kotlin/events/boudicca/eventcollector/collectors/ValugCollector.kt
```diff
@@ -0,0 +1,75 @@
...
+    override fun parseEvent(event: VEvent): Event? {
+        val eventName = event.summary.value
+        val formatter = DateTimeFormatter.ofPattern("yyyyMMdd'T'HHmmss")
+        // get by city name to account for daylight saving time
+        val viennaZoneId = TimeZone.getTimeZone("Europe/Vienna").toZoneId()
+        val eventStartDate = LocalDateTime.parse(event.startDate.value, formatter)
+            .atZone(viennaZoneId).toOffsetDateTime()
+        val now = LocalDateTime.now().atZone(viennaZoneId).toOffsetDateTime()
+
+        if (eventStartDate.isBefore(now)) {
```
kadhonn, line 41: by default only future events are shown on the website anyway, so you can leave or remove this check, however you like

kadhonn:
hey, awesome, thanks for adding a new eventcollector!

i left some small comments, please have a look at them.

for your question about "accessibility.accessibleEntry", absence means "we don't know" and false means no. i also have entered this information into the google sheet https://docs.google.com/spreadsheets/d/1yYOE5gRR6gjNBim7hwEe3__fXoRAMtREkYbs-lsn7uM/edit#gid=0
please read my other comment for more information about that

kadhonn (changes requested):

boudicca.events/eventcollectors/src/main/resources/ical4j.properties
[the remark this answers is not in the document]
dmorawetz: There was a log line complaining about this missing configuration file. Even though it is empty, adding it removed the warning.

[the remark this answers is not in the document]
kadhonn: that is weird, a warning from ical4j? :D but ok sure, if that removes the warning it is fine!

boudicca.events/eventcollectors/src/main/kotlin/events/boudicca/eventcollector/collectors/ValugCollector.kt
dmorawetz: Great, thank you. I wondered how this would be done, but have not found it in the documentation.

Please add an alias for the location, since the location in the ical is mostly written as `VALUG-Klubraum, Alter Schl8hof Wels`.

Furthermore, the coordinates seem to be in the wrong order. The [type definition](https://github.com/boudicca-events/boudicca.events/blob/main/SEMANTIC_CONVENTIONS.md#data-types) states `<longitute>, <latitude>`, and [OSM](https://www.openstreetmap.org/api/0.6/node/1148138856) says the value starting with 14 is the longitude, but when I compare it to the rest of the entries (especially the other entry for schlachthof) it seems to be in the wrong order.

kadhonn: i added the alias

and whoops, thanks for noticing!
i think i messed up the order in the docs, i think normally the default is latitude and then longitude, at least my googling shows me mainly this order :D

kadhonn (approved):

kadhonn:
looks great now, thanks again for your contribution!
but i will squash the changes when merging, hope that is ok with you :D 

also, a short question to you
i took away that i still have not documented the google sheet and that the doc for the coordinates is switched and i will fix them in the evening

disregarding those two issues, how difficult was it to create the eventcollector? was the documentation ok? did you miss something which should be better described? could you test the collector locally without problems?

dmorawetz:
Hey, thank you so much for your quick feedback and merging.

The software architecture is pretty straightforward. The .run folder for local debugging is remarkable! For me, it was literally a clone-and-run experience. It was almost too straightforward, in the sense, that I just looked at other collectors and missed the info about the TwoStepEventCollector in DEV.md. When I finally read on, things made more sense^^

I think the SEMANTIC_CONVENTIONS.md could be linked from DEV.md, REST.md or the README. It is pretty central to know the available fields. 

Great developer experience overall :)

kadhonn:
Awesome, thanks for the feedback!
and yes, linking the semantic values should have been an obviois thing to do, i will add it :D

PR #331, opened 2024-03-08
Abl/remote collector

kadhonn:
first step to what we discussed last time, i made an interface + client for an remote collector. please take a look if you are happy with the design or if you want to change something.

i have created a sample remote-collector you can find here: https://github.com/boudicca-events/remote-collector-samples
also, to use and test this sample + remote collector you can use the LocalCollectorDebug.kt file and use the line 
`.debug(RemoteCollectorCollector("http://localhost:8080", "sample"))`

The important interface is https://github.com/boudicca-events/boudicca.events/pull/331/files#diff-4688436d001ba88c1314e2af6738501490474eebec14d96204465b973c5a04e7

Also, there are quite some changes because i had to adapt the monitoring data structure to be able to use the remote data for it. I recommend reviewing single commits, that should be more obvious what change is what


boudicca.base/eventcollector-client/src/main/kotlin/base/boudicca/api/eventcollector/EventCollectorDebugger.kt
```diff
@@ -24,19 +23,14 @@ class EventCollectorDebugger {
         collectedEvents.forEach {
             println(it)
         }
-        println(Collections.getAllPastCollections()[0])
+        val fullCollection = Collections.getAllPastCollections()[0]
```
twatzl, line 26: is this basically equivalent to Collections.getLastCollection() ?

boudicca.base/eventcollector-client/src/main/kotlin/base/boudicca/api/eventcollector/collections/CollectionClasses.kt
```diff
@@ -9,43 +8,73 @@ data class FullCollection(
...
     override fun toString(): String {
-        return "SingleCollection(\nid=$id, \nstartTime=$startTime, \nendTime=$endTime, \ncollector=$collector, \nhttpCalls=$httpCalls, \nlogLines=${
-            logLines.map {
-                Pair(
-                    it.first,
-                    String(it.second)
-                )
-            }
-        })"
+        return "SingleCollection(\nid=$id, \ncollectorName='$collectorName', \nstartTime=$startTime, \nendTime=$endTime, \ntotalEventsCollected=$totalEventsCollected, \nhttpCalls=$httpCalls, \nlogLines=$logLines, \nerrorCount=$errorCount, \nwarningCount=$warningCount)"
```
twatzl, line 76: jfyi Kotlin can do multiline strings which would make something like this more readable 
https://kotlinlang.org/docs/strings.html#multiline-strings

boudicca.base/remote-collector/remote-collector-api/src/main/kotlin/base/boudicca/api/remotecollector/RemoteCollectorApi.kt
```diff
@@ -0,0 +1,28 @@
...
+import io.swagger.v3.oas.annotations.OpenAPIDefinition
+import io.swagger.v3.oas.annotations.Operation
+import io.swagger.v3.oas.annotations.responses.ApiResponse
+import javax.ws.rs.GET
+import javax.ws.rs.POST
+import javax.ws.rs.Path
+import javax.ws.rs.Produces
+
+@OpenAPIDefinition
+interface RemoteCollectorApi {
```
twatzl, line 13: So i guess we do not see this api as on the same level as the publisher api?

boudicca.base/eventcollector-client/src/main/kotlin/base/boudicca/api/eventcollector/collectors/RemoteCollectorCollector.kt
```diff
@@ -0,0 +1,46 @@
+package base.boudicca.api.eventcollector.collectors
+
+import base.boudicca.api.eventcollector.EventCollector
+import base.boudicca.api.eventcollector.collections.Collections
+import base.boudicca.api.remotecollector.RemoteCollectorClient
+import base.boudicca.api.remotecollector.model.HttpCall
+import base.boudicca.model.Event
+
+class RemoteCollectorCollector(private val url: String, private val name: String? = null) : EventCollector {
```
twatzl, line 9: maybe it would be good to add a short javadoc explanation what this one does.

boudicca.base/remote-collector/remote-collector-api/src/main/kotlin/base/boudicca/api/remotecollector/RemoteCollectorApi.kt
kadhonn: why do think this way? because i moved it into a subfolder? that was only done to group it better and we already have a looot of projects in the base.boudicca folder

boudicca.base/eventcollector-client/src/main/kotlin/base/boudicca/api/eventcollector/collections/CollectionClasses.kt
kadhonn: hm... probably, but this is was my intellij generated, so I didn't want to change it :D

boudicca.base/eventcollector-client/src/main/kotlin/base/boudicca/api/eventcollector/EventCollectorDebugger.kt
kadhonn: kinda, but it would make sense to add a helper method for this, i will do this

boudicca.base/remote-collector/remote-collector-api/src/main/kotlin/base/boudicca/api/remotecollector/RemoteCollectorApi.kt
twatzl: Just because it was in a subfolder and hence different from the other apis. but grouping makes sense to me.

twatzl (approved):

PR #341, opened 2024-03-14
extend semantic keys and improve documentation

twatzl:


SEMANTIC_CONVENTIONS.md
```diff
@@ -28,33 +28,106 @@ We use certain data types for the properties we expect.
...
+#### tags values
+
+tags is an open field, so any value is permitted, but this is a list of common/known tags:
+
+- ...
+- TODO
+
+#### Image json
+
+```json
```
kadhonn, line 113: we really have to talk about this change, because this changes our fundamental data model @Yolgie

SEMANTIC_CONVENTIONS.md
```diff
@@ -28,33 +28,106 @@ We use certain data types for the properties we expect.
...
+- workshop
+- ...
+
 #### Category enum values
 
 * `MUSIC`: concerts or other events where the main focus is music
 * `TECH`: event with technology as the focus
 * `ART`: art exhibitions, comedy, theater, ...
 
+#### RecurrenceType enum values
```
kadhonn, line 86: i would be fine with those enum values, even though i would not need the distinction between rarely and regularly

SEMANTIC_CONVENTIONS.md
```diff
@@ -15,7 +15,7 @@ whole JSON-document should be encoded in UTF-8.
 
 We use certain data types for the properties we expect.
 
-* `text`: just a simple text/string
+* `text`: just a simple text/string*
```
kadhonn, line 18: why is the star added here?

SEMANTIC_CONVENTIONS.md
```diff
@@ -28,33 +28,106 @@ We use certain data types for the properties we expect.
...
+| recurrence.interval  | Describing the interval with which the event recurs                                                  | text                                 | OPTIONAL    |
+| tags                 | A list of tags. TODO how to describe?                                                                | list\<text>                          | RECOMMENDED |
+| registration         | If this is a free event, a event which requires registration or a event which requires a paid ticket | enum\<registration>                  | OPTIONAL    |
+| pictureUrl           | Url to a picture to be shown                                                                         | url                                  | DEPRECATED  |
+| pictures.json        | A json array of pictures for the event. The first picture should be the featured one.                | json\<image>                         | OPTIONAL    |
+| collectorName        | Name of the collector which collected this event                                                     | text                                 | MANDATORY   |
```
kadhonn, line 47: not sure if we should make this mandatory, currently it is not

SEMANTIC_CONVENTIONS.md
```diff
@@ -28,33 +28,106 @@ We use certain data types for the properties we expect.
...
+| sources              | A list of all sources, line by line. This should include all URLs or other sources used.             | list\<text>                          | RECOMMENDED |
+| source.event.page    | the url of the page where the event was sourced from                                                 | url                                  | RECOMMENDED |
```
kadhonn, line 49: not sure for what we would use those values?

SEMANTIC_CONVENTIONS.md
twatzl: ok. I just documented the current status. I thought it was mandatory, but then I will change it.

kadhonn:
i was thinking a lot about this in the last few days, because something just was not feeling right for me and i cannot just let that be.... anyway, one attempt to sort that out for me:

maybe we should split our semantic conventions?
that sounds stupid at first, but hear me out
first, we always wanted to have a less strict data-model, because there are already a lot of other strict data models, which is one reason we exist now :D

but that means that different instances of boudicca, especially if they have a bit of a different focus like museumsbahnen has, will always have different requirements and will always use custom keys with probably custom formats?. i did that too already with a few different things to try it out.
why don't we make this official by documenting that we accept different ways/different standards? 
of course that means everyone needs to document their standard which maybe make custom keys more visible?
we then would have one for our boudicca.events website and museumsbahnen will have a bit different one

now the problem is of course interoperability
there are some ways we could tackle that. maybe you just have to bite the bullet and write a custom converter for all boudicca instances you want to ingest data from?
or maybe we could have some boudicca "canonical" format, and each eventdb has the current normal export and an "canonical" export, which maps all data to the boudicca canonical format. then everyone can import from that format and map it to his.

buuut maybe this all is stupid and we should really work on one defined standard, i still don't know :D
do you have any thoughts about this?

also, maybe this belongs in it's own discussion/issue :D

twatzl:
I already have this and I think it makes totally sense for things that are specific for museum railways. 

however something like event images and sources are universal across all events.

This is why i created the MR here.

https://github.com/museumsbahn-it-community/museumsbahn-events/blob/main/museum-railway-api/src/main/kotlin/at/museumrailwayevents/model/conventions/CommonKeys.kt

kadhonn (approved):

PR #360, opened 2024-03-24
add museumsbahnen

kadhonn:


Yolgie (approved):

kadhonn:
one issue blocking this is we need to show the copyright property of an image for those events

Yolgie:
this should be unblocked now right @twatzl ?

PR #373, opened 2024-04-05
comply with image-alt and region rules of a11y

### LINKED ISSUE #369: Playwright report issues generated on `https://boudicca.events` (deployed version)
basiclly after analyzing the report generated by automatic a11y violation detector, there are two rules which are failing:

1. image-alt
2. region

in the following comments, these two are analyzed throughly.

mahdikhashan:
resolves #369


boudicca.base/publisher-event-html/src/main/resources/templates/events/event.hbs
```diff
@@ -1,7 +1,7 @@
 <div class="event">
         <div class="event-image {{#if event.category}}event-image-{{event.category}}{{/if}}">
             {{#if event.pictureUuid}}
-                <img src="/picture?uuid={{event.pictureUuid}}" style="height:100px;max-height:250px;max-width:300px;" />
+                <img src="/picture?uuid={{event.pictureUuid}}" style="height:100px;max-height:250px;max-width:300px;" alt="{{event.name}} bild" aria-label="Event Bild" aria-labelledby="image" />
```
kadhonn, line 4: we have a new defined property called pictureAlt described here: https://github.com/boudicca-events/boudicca.events/blob/main/SEMANTIC_CONVENTIONS.md#L47

so we should try to actually use it

so please have a look if this property exists for an event, and if so you use that. if it does not exist use the text you are using right now as a fallback

kadhonn, line 4: are we sure that alt and aria-label should be different from each other?

mahdikhashan: I was experimenting tags, we should only keep alt, since it's an `img` tag itself and represents it's role. but, with the current screen reader extension, I was unable to validate neither. non is read by the screen reader.

mahdikhashan: will check it out and use it.

mahdikhashan: done

kadhonn (approved):
please squash the commits before or when merging, otherwise looks great!

PR #400, opened 2024-04-19
modal event details

mahdikhashan:


boudicca.base/publisher-event-html/src/main/resources/static/js/main.js
```diff
@@ -10,6 +10,64 @@ document.addEventListener("DOMContentLoaded", () => {
...
+  const modalContent = modal.querySelector('.modal-content');
+
+  events.forEach(event => {
+    event.addEventListener('click', () => {
+
+      const eventAttributes = event.dataset;
+
+      modalContent.innerHTML = `
+        <div class="modal-flex-container">
+            <div class="modal-column">
```
kadhonn, line 26: i don't like the html in javascript way
i think it would be way better to render each full event as it should be shown in the end, but all of them are hidden
then when you expand it simply make the correct one visible

boudicca.base/publisher-event-html/src/main/resources/templates/events/event.hbs
```diff
@@ -1,48 +1,63 @@
...
-            {{else}}
-                <svg title="Event Bild" viewBox="0 0 512 512" height="100px" >
-                    <use xlink:href="#{{#if event.category}}{{event.category}}{{else}}image{{/if}}"></use>
+<div class="event"
+     data-category="{{event.category}}"
+     data-picture-uuid="{{event.pictureUuid}}"
+     data-picture-alt-text="{{event.pictureAltText}}"
+     data-url="{{event.url}}"
+     data-accessibility-properties="{{event.accessibilityProperties}}"
+     data-name="{{event.name}}"
```
kadhonn, line 7: if we prerender the modal/drawer content we also don't need to do this

boudicca.base/publisher-event-html/src/main/kotlin/base/boudicca/publisher/event/html/service/EventService.kt
```diff
@@ -188,12 +189,12 @@ class EventService @Autowired constructor(
             }
             if (category != null) {
                 return when (category) {
-                    EventCategory.MUSIC -> "music"
-                    EventCategory.ART -> "miscArt"
-                    EventCategory.TECH -> "tech"
-                    EventCategory.SPORT -> "sport"
-                    EventCategory.ALL -> "???"
-                    EventCategory.OTHER -> null
+                    EventCategory.MUSIC -> "Musik"
```
kadhonn, line 192: those values are used for the category fallback icons, those are not normal text. so by changing them you broke all the icons :)

boudicca.base/publisher-event-html/src/main/resources/static/js/main.js
```diff
@@ -9,6 +9,66 @@ document.addEventListener("DOMContentLoaded", () => {
...
+      const detailButton = event.querySelector('.anchor-to-event')
+      const closeButton = event.querySelector(".modal-close-button")
+
+      modal.addEventListener('click', () => {
+        if (event.target === modalContent) {
+          return;
+        }
+
+        modal.style.display = "none";
+        document.body.style.overflow = 'initial';
```
kadhonn, line 27: those two lines make excellent candidates to extract into their own method, called something like "closeModal()" or "openModal()"
that has the benefit of:
1) making clear what those lines are supposed to to, namely open or close the modal
2) de-duplicates those lines of code. currently they are duplicated multiple times here which makes it harder to change without introducing mistakes

boudicca.base/publisher-event-html/src/main/resources/static/js/main.js
```diff
@@ -107,6 +167,13 @@ document.addEventListener("DOMContentLoaded", () => {
       onSearchLoadMoreButtonBehaviour(ssrDomEventString);
       const newEvents = parser.parseFromString(ssrDomEventString, "text/html");
       eventsContainer.append(...newEvents.body.children);
+
+      ////
```
kadhonn, line 171: what are those //// supposed to mean?

boudicca.base/publisher-event-html/src/main/resources/static/js/main.js
```diff
@@ -9,6 +9,66 @@ document.addEventListener("DOMContentLoaded", () => {
   const loadMoreButton = document.getElementById("loadMoreButton");
   const categorySelect = document.getElementById("category");
   const categoryFieldSets = document.querySelectorAll("[data-category-wanted]");
+  const events = document.querySelectorAll('.event');
```
kadhonn, line 12: why do you read all events here? this events-list will be out of date pretty with the next search/load more button and does not seem to get updated?

boudicca.base/publisher-event-html/src/main/resources/static/js/main.js
```diff
@@ -107,6 +167,13 @@ document.addEventListener("DOMContentLoaded", () => {
       onSearchLoadMoreButtonBehaviour(ssrDomEventString);
       const newEvents = parser.parseFromString(ssrDomEventString, "text/html");
       eventsContainer.append(...newEvents.body.children);
+
+      ////
+      const newlyAddedEvents = eventsContainer.querySelectorAll('.event');
+      setModalBehaviour(newlyAddedEvents)
```
kadhonn, line 173: with this you re-add all event listeners to all events. so after a load-more, the original events have all listeners applied to them twice

boudicca.base/publisher-event-html/src/main/resources/templates/events/event.hbs
```diff
@@ -1,48 +1,439 @@
+<style>
+    .event {
```
kadhonn, line 2: please move this back to events.hbs, because event.hbs is rendered in a loop, so you now have this css block for each event we show :D

kadhonn (changes requested):

boudicca.base/publisher-event-html/src/main/resources/templates/events/event.hbs
mahdikhashan: yep, done.

boudicca.base/publisher-event-html/src/main/resources/static/js/main.js
mahdikhashan: removed.

boudicca.base/publisher-event-html/src/main/kotlin/base/boudicca/publisher/event/html/service/EventService.kt
mahdikhashan: It has been fixed.

boudicca.base/publisher-event-html/src/main/resources/static/js/main.js
mahdikhashan: done.

boudicca.base/publisher-event-html/src/main/resources/static/js/main.js
mahdikhashan: done

boudicca.base/publisher-event-html/src/main/resources/static/js/main.js
mahdikhashan: I couldn't find a better way, would it be possible to check it and give me feedback?

boudicca.base/publisher-event-html/src/main/resources/static/js/main.js
```diff
@@ -10,6 +10,34 @@ document.addEventListener("DOMContentLoaded", () => {
   const categorySelect = document.getElementById("category");
   const categoryFieldSets = document.querySelectorAll("[data-category-wanted]");
   const searchInput = document.querySelector("input.search-input");
+  const modal = document.getElementById("modal");
+  const modalContent = modal.querySelector("#modal-content");
```
kadhonn, line 14: why do you use the querySelector method with ids instead of the normal getElementById?

boudicca.base/publisher-event-html/src/main/resources/static/js/main.js
```diff
@@ -161,17 +195,26 @@ document.addEventListener("DOMContentLoaded", () => {
...
     } catch (e) {
       console.error(e);
     }
   };
 
+  const initModals = (events) => {
+    events.forEach(event => {
+      const anchor = event.querySelector(".anchor-to-event");
+      const content = event.querySelector("#modal-content");
+      anchor.addEventListener("click", () => {
```
kadhonn, line 212: i am pretty sure you add this event listener to already existing events as well, so events can get multiple event listeners when you click "load more"

boudicca.base/publisher-event-html/src/main/resources/templates/events/event.hbs
```diff
@@ -8,14 +8,14 @@
             </svg>
         {{/if}}
         {{#if event.url}}
-            <a class="anchor-to-eventpage" href="{{event.url}}" target="_blank" aria-describedby="Zur Eventseite von {{event.name}} (neues Fenster wird geöffnet)">
-                Zur Eventseite
-            </a>
+            <button class="anchor-to-event">
+                Weitere Einzelheiten
```
kadhonn, line 12: Maybe we should use "Mehr Informationen" as text here

mahdikhashan: yes, I agree.

mahdikhashan: done.

boudicca.base/publisher-event-html/src/main/resources/static/js/main.js
mahdikhashan: I think there is sth with async code that I don't know how to handle, when I access the modal-content by document. getElementById, I'm not able to fill it with the content later. however, I changed the selector for the close-button as was requested.

kadhonn (approved):
approved but tests are still failing

kadhonn (approved):

PR #407, opened 2024-04-23
structured data adr

### LINKED ISSUE #367: Structured Data
In the last few weeks we had multiple discussions about different features that need more structure in our simple string->string map, mainly discussions about lists, i18n features and embedding more structure like json for combining lists and objects or similar stuff.
But we never got a real satisfactory answer to how to do it in a good way, so to start this discussion here are my thoughts and reasoning until now..

# Structured Data

## Motivation

So, why/where do we need to care about the different data formats/structures?

Our Architecture is structured in three parts: Collectors, Core, Publishers
Collectors and Publishers will mostly be specialized and to some degree have to agree on the same property names/format anyway to work together, so this is not the primary concern here. But our Core, especially the Search Service is conceptually as generic as possible, meaning it should work with pretty much all property names/formats. And for things like searching, sorting, faceting, ... to work correctly, we need to know how to handle the different data structures.

Some features we have do need more structure to be supported fully, I have identified some for now:

1. **Lists.** Sometimes a property will have more then one value, for example in our `concert.bandlist` property
2. **Variants.** Properties can have multiple variants of it's value, for example i18n properties which have different translations
3. **Dates.** We have dates in our data, `startDate` and `endDate` for now, but there may be others as well. This is important for sorting
4. (Maybe?) **Numbers.** Maybe we want to support numbers specifically? Also important for sorting
5. (Maybe?) **Structured Data.** This one is kinda weird and could be anything? I am not sure how to generalize this. Maybe we should call this blob or something as a marker to ignore normal handling and treat it as opaque?

## Possible solutions

I can think of three ways to convey structure:

1. **Hardcode it.** Simply, the search service has to know beforehand how data is to be handled, via code or some configuration files. This would be about the same data as the Collectors and Publishers have as well and would make it not generic at all. At this point we simply have a normal schema, so we could just as well use ical or any other standard instead of making our own, so I do not prefer this.
2. **Guess it.** Look at the data and/or query and try to guess what kind of data it has to be. This is what currently is done and it is errorprone because sometimes there can be overlaps in structure, like a normal text beginning with a "[" which could be interpreted as a list.
3. **Metadata.** We could also add some metadata to the event itself, which publishers and core services could look at. Examples would be something like having the normal property `name` and then having a meta-property named `?name` or `name$schema` which contains the structure of the data, maybe a simple `date` or `list`. Or another example would be what Open Street Map is doing with its i18n support by adding some metadata to the property name itself, like `name:cz` and `name:en` for different locales. You could do this for lists as well. We of course could embed this metadata in the value of the property itself, but that makes everything a bit more complicated to parse, especially because you would need special handling for simple text as well.

Soo, that is how far I am with this topic, what are your thoughts?

kadhonn:
closes #367


docs/adr/ADR-01-STRUCTURED_DATA.md
```diff
@@ -0,0 +1,148 @@
...
+* Makes key-handling more complex, since there are now multiple keys for one "base-key"
+
+
+### Decision
+
+We decided on putting the format and variant information into the key, as different key=value pairs seperated by ":"
+
+At this point we need to introduce some more precise terminology:
+
+TODO move or copy this glossary somewhere else?
```
twatzl, line 73: would move the final documentation to SEMANTIC_CONVENTIONS.md

docs/adr/ADR-01-STRUCTURED_DATA.md
```diff
@@ -0,0 +1,148 @@
...
+* (an empty list)
+* value1,,value2 (this is a list consisting of three values "value1", "" (an empty value) and "value2")
+* value1, value2 (this is a list consisting of two values "value1" and " value2" (please note the preceding whitespace in the second value))
+* , (this is a list consisting of two empty values, please note that there is no way to specify a list with one empty value)
+
+### Markdown
+
+Knowledge about markdown is important for rendering it correctly.
+
+The format is normal [Markdown](https://en.wikipedia.org/wiki/Markdown).
```
twatzl, line 148: I would use a different wording. 

e.g. "We agree to use standard markdown according to wiki"

docs/adr/ADR-01-STRUCTURED_DATA.md
```diff
@@ -0,0 +1,148 @@
...
+
+The format is the "Date and time with the offset" format of the [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) standard.
+
+Examples:
+* 2024-04-27T23:59:00+02:00
+* 2024-04-27T11:00:00Z
+
+### Lists
+
+Knowledge about lists is mostly important for contains searches and search faceting. We cannot sort lists. (TODO can/should we?)
```
twatzl, line 131: sorting according to list values should be done by client, not?
it would be more complex as we would also need some information about the sort order. maybe we have to implement this later when we find a use case for it in the publisher, but for now i would not

docs/adr/ADR-01-STRUCTURED_DATA.md
```diff
@@ -0,0 +1,148 @@
...
+* Numbers (VariantValue "number")
+* Dates (VariantValue "date")
+* Lists (VariantValue "list")
+* Markdown (VariantValue "markdown")
+
+### Numbers
+
+Knowledge about numbers is mostly important to correctly sort them.
+
+The format looks like: TODO define format and possible numbers here
```
twatzl, line 111: format is just javascript number... or not?

docs/adr/ADR-01-STRUCTURED_DATA.md
```diff
@@ -0,0 +1,148 @@
...
+
+Currently supported variants are:
+* Language: which language a value is in
+* Format: what kind of format the value is in
+
+If a certain variant makes sense for a certain property needs to be decided per property basis.
+
+#### Language Variant
+
+Has the VariantName "lang" and the value is a two-letter language abbreviation, like "en", "de", "fr", ...
```
twatzl, line 88: how would we denote the "original" language?
Or do we use some kind of indication of collected value vs enriched value?

docs/adr/ADR-01-STRUCTURED_DATA.md
kadhonn: i would like it to be more specific, what do we want to support? only "normal" numbers like 2 and -3? floating point? hex values? e notation numbers?

docs/adr/ADR-01-STRUCTURED_DATA.md
kadhonn: hmm
maybe an event already has two languages on it's source?
i think we should just annotate generated values extra, the rest should be as is
we could think of having no lang variant as the default?

docs/DATA_MODEL.md
```diff
@@ -0,0 +1,299 @@
...
+One example would be `description:lang=de`, which is a filter filtering for the PropertyName `description` and the
+Variant `lang=de`.
+This is done by taking all available PropertyVariants for the Property `name` and then only taking those whose `lang`
+Variant matches the value `de`. It does not matter if the PropertyVariant has additional Variants, those will be
+ignored.
+
+If KeyFilters return more than one PropertyVariant, the result will be sorted alphabetically by
+their [Canonical Form](#canonical-form).
+
+One difference compared to Keys is the empty VariantValue, like `lang=`. This means "select all PropertyVariants which
```
p4dd9, line 120: hm ok!

docs/DATA_MODEL.md
```diff
@@ -0,0 +1,299 @@
...
+do NOT have the VariantName at all".
+Another difference is that `*` is allowed as a PropertyName, this would match and select all Properties. You can still
+add Variant Filters to the `*` filter.
+
+Let's look at some examples given following PropertyVariants of the `description` Property:
```
p4dd9, line 125: i love the examples!

docs/DATA_MODEL.md
```diff
@@ -0,0 +1,299 @@
...
+
+With the filter `description` you would get all PropertyVariants as a result, because there is no filter except the
+PropertyName.
+
+With the empty VariantValue filter `description:lang=` you will only get `description` as a result, since this is the
+only PropertyVariant without the `lang` Variant.
+
+## Unknown Variants
+
+Boudicca will ignore all Variants it does not understand and treat the keys as if those Variants are not here. This
```
p4dd9, line 144: [nit[ as a noob, i might understand why it does not work - http statuscode and/or message on requests level would be beneficial to avoid wasting some more time finding out what the issue is

p4dd9: Hm, just thinking out loud - either ignore it - or return no results at all. I totally think this/current approach is totally fine

docs/DATA_MODEL.md
```diff
@@ -0,0 +1,299 @@
...
+behaviour is suggested for all Software handling Boudicca Keys.
+
+## Variants and the Search Service
+
+The Search Service will evaluate [Queries](QUERY.md) like the `contains` Query by searching all PropertyVariants
+matching the [KeyFilter](#keyfilters) given to the Query and then performing the Query against all the PropertyVariants.
+If any match, it will evaluate to true.
+The only Variant the Search Service does know about is the [Format Variant](#format-variant), because this will
+influence how searching. parsing and ordering works on those Variants.
```
p4dd9, line 153: ,

docs/DATA_MODEL.md
```diff
@@ -0,0 +1,299 @@
...
+Variant, which leaves us with the PropertyVariants `description`, `description:lang=de` and `description:lang=en`. It
+will now evaluate the `contains` query against the value of those three PropertyVariants, and if any match the query
+will evaluate to true.
+
+## Variants and the HTML-Publisher
+
+For displaying properties with different Formats the HTML-Publisher follows
+the [Semantic Conventions](SEMANTIC_CONVENTIONS.md) and tries to find the Variant with the highest priority.
+
+If the user has selected a language (feature pending :P ) the Publisher will preferably select a PropertyVariant with
```
p4dd9, line 177: [nit] hightlight this one if language select is visible in ui

docs/DATA_MODEL.md
```diff
@@ -0,0 +1,299 @@
...
+* `description:lang=en`
+* `description:lang=en:format=date`
+
+For the query `"description" contains "my search text"` Search Service will first select all PropertyVariants which
+match the given KeyFilter `description`, which are all of the above. Then it will discard all with an invalid format
+Variant, which leaves us with the PropertyVariants `description`, `description:lang=de` and `description:lang=en`. It
+will now evaluate the `contains` query against the value of those three PropertyVariants, and if any match the query
+will evaluate to true.
+
+## Variants and the HTML-Publisher
```
p4dd9, line 172: what is an html publisher?

docs/DATA_MODEL.md
```diff
@@ -0,0 +1,299 @@
...
+## Variants and the iCal-Publisher
+
+Similar to the HTML-Publisher, only that it cannot render Markdown and will ignore this format.
+
+# Known Variants
+
+## Language Variant
+
+Has the VariantName "lang" and the value is a two-letter language abbreviation, like "en", "de", "fr", ...
+Denotes the language the value is in. Without a language variant it is assumed that this is the "default" language and
```
p4dd9, line 215: what is the default lang as fallback? can I request multiple lang variantNames chained via ","?

docs/DATA_MODEL.md
```diff
@@ -0,0 +1,299 @@
...
+will be used as a fallback if the wanted language is not available.
+
+## Format Variant
+
+Has the VariantName "format" and specifies what kind of data and in which format the Value is.
+
+Boudicca currently supports those formats:
+
+* Text (the default when no format is given, please do NOT use any VariantValue for this)
```
p4dd9, line 224: highlight

docs/DATA_MODEL.md
```diff
@@ -0,0 +1,299 @@
...
+* 0.5
+
+Sorting: By numerical value
+
+Searching: Works with Equals queries. Numbers need to be equal. TODO we probably need greater and smaller-queries here.
+
+### Dates
+
+Format: The format is the "Date and time with the offset" format of
+the [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) standard. TODO not sure if we really want to accept all ISO
```
p4dd9, line 256: highlight light todo for possible upcoming "breaking" changes

docs/DATA_MODEL.md
```diff
@@ -0,0 +1,299 @@
...
+formats, i guess there are local ones in there as well (ones without timezone information)
+
+Examples:
+
+* 2024-04-27T23:59:00+02:00
+* 2024-04-27T11:00:00Z
+
+Sorting: By time.
```
p4dd9, line 264: so the total time in milis? datetime?

docs/QUERY.md
```diff
@@ -4,81 +4,58 @@
...
+  descriptions: `"description:lang=en" contains "technology"`
+* Search for any event NOT containing the text "technology" in the name: `not ("name" contains "technology")` (parenthesis are not needed in this example but make it clearer)
 * Search for any event containing the text "technology" in any field (like when you search on our
   website): `"*" contains "technology"`
 * Search for any event in either Linz or Enns: `"location.city" equals "Linz" or "location.city" equals "Enns"`
 * Search for any event containing `a "longer" sentence` in the
   description: `"description" contains "a \"longer\" sentence"`
 * Search for any event in Wien but not in
-  Gasometer: `"location.city" equals "Wien" and (not "location.name" equals "Gasometer")`
+  Gasometer: `"location.city" equals "Wien" and (not "location.name" equals "Gasometer")` (parenthesis are not needed in this example but make it clearer)
```
p4dd9, line 55: highlight notes

docs/SEMANTIC_CONVENTIONS.md
```diff
@@ -0,0 +1,127 @@
...
+* `enum<?> (format=text)`: Has to be one of the specified distinct values
+* `boolean (format=text)`: The text "true" or "false"
+
+Note: publishers have to be able to handle invalid values by for example simply ignoring them
+
+### Property List
+
+**Events have only two required properties: `name` and `startDate`**
+
+| PropertyName                                   | Meaning                                                                                                                         | [Data Type](#data-types) | [Usage](#usage-explanation) |
```
p4dd9, line 30: how do you plan to keep this list up-2-date?

docs/SEMANTIC_CONVENTIONS.md
```diff
@@ -0,0 +1,127 @@
...
+- guided_tour
+- workshop
+- ...
+
+#### "tags" property
+
+tags is an open field, so any value is permitted, but this is a list of common/known tags:
+
+- ...
+- TODO
```
p4dd9, line 97: todo

p4dd9 (approved):
Nice!

docs/DATA_MODEL.md
```diff
@@ -0,0 +1,299 @@
+# Boudicca's Data Model
+
+## Overview
+
+Boudicca is a database containing Entries, which are a collection of `Key`->`Value` Pairs, so-called Properties.
```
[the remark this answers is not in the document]
kadhonn: an entry is nothing specific, it has no meaning. meaning is given by treating it as an event for example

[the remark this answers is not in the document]
kadhonn: i moved the section explaining this up a bit so it will come right after this sentence

docs/DATA_MODEL.md
```diff
@@ -0,0 +1,299 @@
...
+For example `name`->`My Entry` would be a Property with key "name" and value "My Entry" and could represent the name of
+this Entry.
+
+Sometimes you need more information about the value of one property, for example which language the value is in, or if
+this value is supposed to be a number, text, date, ...
+Boudicca encodes this information into the Key of the Property by appending Variants to the property name, which look
+like `:VariantName=VariantValue`
+For example to specify that this name is in english we could have the Property `name:lang=en`->`My Entry` or to specify
```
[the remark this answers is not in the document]
kadhonn: thanks, i will document this list/standard on the language variant secion

docs/DATA_MODEL.md
```diff
@@ -0,0 +1,299 @@
...
+that something is a number we can write `price:format=number`->`20`.
+See [Format](#format-variant) and [Language](#language-variant) for more information on those Variants.
+You can of course chain and combine those Variants, but be aware that most of the time the same Variant is only allowed
+to appear once per Key and watch out for the [canonical form](#canonical-form).
+
+An Entry is a very generic way of viewing the data in Boudicca and most of the time you will only think about and work
```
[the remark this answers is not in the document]
kadhonn: i reworded the whole sentence a bit, hopefully it makes more sense now

docs/DATA_MODEL.md
```diff
@@ -0,0 +1,299 @@
...
+* Event: A specialised `Entry`, which represents an event and has two mandatory keys: `name` and `startDate`.
+  See [Semantic Conventions](SEMANTIC_CONVENTIONS.md) for more information.
+* Property: All `Key`->`Value` pairs where the key has the same `PropertyName` but different `Variants`.
+* PropertyVariant: One `Key`->`Value` pair of a property.
+* Key: Key value of one `Key`->`Value` mapping. Can contain multiple `Variant`. Format is `PropertyName[:Variant]*` and
+  it is encoded as a UTF-8 string.
+* Value: The value of one `PropertyVariant`. Has to conform to the specified `Variant`, so conform to the correct
+  language and/or format. It is encoded as a UTF-8 string.
+* PropertyName: Name of one `Property` of an `Entry`. Can only consist of letters, numbers and `.`
+* Variant: Specifying what kind of Variant this key is. Format is `VariantName=VariantValue`
```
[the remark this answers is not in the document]
kadhonn: added examples to this

docs/DATA_MODEL.md
kadhonn: ?

docs/DATA_MODEL.md
kadhonn: well, the reason i want to ignore and not reject unknown variants is that maybe a user wants to add some new variant to his data. it also makes it easier to introduce new known variants because old clients will still work

docs/DATA_MODEL.md
kadhonn: what do you mean?

docs/DATA_MODEL.md
kadhonn: that is our website

docs/DATA_MODEL.md
kadhonn: "what is the default lang as fallback?" it is not a language, it is meant as a default text compared to variants who explicitly set a language.

docs/DATA_MODEL.md
kadhonn: highlight what and how?

docs/DATA_MODEL.md
kadhonn: changed it to Chronologically

docs/DATA_MODEL.md
kadhonn: what exactly do you mean?

docs/SEMANTIC_CONVENTIONS.md
kadhonn: by updating it when we change or introduce new properties :P   not sure what kind of strategy you want here?

docs/SEMANTIC_CONVENTIONS.md
kadhonn: yeah, that todo is open since the beginning of (boudicca) time, and i still don't know better :D

docs/DATA_MODEL.md
p4dd9: in markdown, it should be nice if you add smth that highlights that there will changes via "!" emoji or bold text

docs/SEMANTIC_CONVENTIONS.md
p4dd9: just ooc, manual updates usually run out of sync with code,. if it works, it works

docs/DATA_MODEL.md
p4dd9: AHA!

docs/SEMANTIC_CONVENTIONS.md
kadhonn: yes it does run out of sync, but i really cannot think of a better way right now.

docs/DATA_MODEL.md
kadhonn: yeah, but i just made an decision and disallowed local dates :P

Issue #414, opened 2024-04-24
A11Y: manually test HtmlPublisher wrt to WCAG 2 


mahdikhashan:

mahdikhashan:
latest update on the issue:

1. search field flow fixed, review remains.
2. change requested for focus on the page wrapper still exists.

PR #419, opened 2024-04-29
make the close button the first tabbable element

mahdikhashan:
addresses #414


mahdikhashan:
forward path with `TAB` and backward with `TAB + SHIFT`

https://github.com/boudicca-events/boudicca.events/assets/58775404/7aa9a4f2-285a-45de-b596-c35d54b87aae

kadhonn (approved):

PR #421, opened 2024-04-30
a11y: change page title

mahdikhashan:
addresses part of # 414 

https://www.w3.org/WAI/WCAG21/Understanding/page-titled.html

### DIFF
--- a/boudicca.base/publisher-event-html/src/main/kotlin/base/boudicca/publisher/event/html/controller/StartPageController.kt
+++ b/boudicca.base/publisher-event-html/src/main/kotlin/base/boudicca/publisher/event/html/controller/StartPageController.kt
@@ -14,7 +14,7 @@ import org.springframework.web.servlet.ModelAndView
 @RequestMapping("/")
 class StartPageController @Autowired constructor(private val eventService: EventService) {
 
-    private val PAGE_TITLE = "Boudicca.Events - Open source event collector and publisher software"
+    private val PAGE_TITLE = "Boudicca.Event - find accessible events in Austria"
 
     @GetMapping("/")
     fun getIndex(): ModelAndView {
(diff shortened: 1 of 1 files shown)


boudicca.base/publisher-event-html/src/main/kotlin/base/boudicca/publisher/event/html/controller/StartPageController.kt
```diff
@@ -14,7 +14,7 @@ import org.springframework.web.servlet.ModelAndView
 @RequestMapping("/")
 class StartPageController @Autowired constructor(private val eventService: EventService) {
 
-    private val PAGE_TITLE = "Boudicca"
+    private val PAGE_TITLE = "Boudicca.Events - Open source event collector and publisher software"
```
kadhonn, line 17: hmm
do we want to focus on events or on our software here? i think the former, so i would focus on that

something mentioning more our loval events, or accessibility or something like that

mahdikhashan: Boudicca.event - find accessible events in austria ?

kadhonn: yeah something like that! :)

boudicca.base/publisher-event-html/src/main/kotlin/base/boudicca/publisher/event/html/controller/StartPageController.kt
```diff
@@ -14,7 +14,7 @@ import org.springframework.web.servlet.ModelAndView
 @RequestMapping("/")
 class StartPageController @Autowired constructor(private val eventService: EventService) {
 
-    private val PAGE_TITLE = "Boudicca.Events - Open source event collector and publisher software"
+    private val PAGE_TITLE = "Boudicca.Event - find accessible events in Austria"
```
kadhonn, line 17: the page is still called "Boudicca.Events", but the additional text looks fine for me

mahdikhashan: done

kadhonn (approved):

PR #428, opened 2024-05-09
Fix time zone of fhLUG events

knittl:
Time zones are one huge PITA. The start time of events was erroneously shifted twice, ending up with start times 2 hours too early. ics evnets already have the timezone specified, so we can parse directly to an OffsetTimeZone. For date-only events ("full day"), use the old logic so that they start at midnight, local time.

Thanks!


kadhonn (approved):
haha, yeah timezones are ass :D

knittl:
@kadhonn Is there anything left to do before this PR can be merged?

Also, one other thing: I noticed that events get duplicated boudicaa if they are updated at the source (e.g. the title changes). I would not have expected this, because the UID of the ical event (and in the event properties) is still identical.

kadhonn:
whoops, no everything is alright, i just forgot^^

and yes, that is currently the case with duplications. currently our "primary key" is name+startDate, so when one of those changes we have duplicates.
you are right that for ical events we would already have a proper uid we could use, but that is not the case for most other eventsources we have, so that is why we do not have a proper handling yet.

PR #430, opened 2024-05-13
add header-subtitle

### LINKED ISSUE #299: Implement suggestions for better first time user experience
Part of the result of the User Tests #289 should be concrete improvements for first time users.

Coordinate with your other USCC Fellows to discuss possible features and create tickets for them.

mahdikhashan:
resolves #299

### DIFF
--- a/boudicca.base/publisher-event-html/src/main/resources/templates/layout/header.hbs
+++ b/boudicca.base/publisher-event-html/src/main/resources/templates/layout/header.hbs
@@ -106,11 +106,11 @@
     }
 
     @media only screen and (max-width: 700px) { 
-        h1 {
+        .header-title {
             display: none;
         }
 
-        h3 {
+        .header-subtitle {
             display: none;
         }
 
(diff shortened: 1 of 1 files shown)


boudicca.base/publisher-event-html/src/main/resources/templates/layout/header.hbs
```diff
@@ -94,6 +110,10 @@
             display: none;
         }
 
+        h3 {
+            display: none;
```
kadhonn, line 114: what is this css rule about?

boudicca.base/publisher-event-html/src/main/resources/templates/layout/header.hbs
```diff
@@ -139,6 +159,7 @@
                 <img src="/img/logo_small.png" alt="Boudicca Logo - Ein rundes Schwarz-auf-Weiß-Bild. Ein stilisiertes Logo der Mondkrone des keltischen Volksstammes Iceni." height="56px" width="56px"/>
                 <div>
                     <h1 class="header-title">{{title}}</h1>
+                    <h3 class="header-subtitle">All accessible events in Austria</h3>
```
kadhonn, line 162: 1) While I agree we should add an english version of our website soonish i think we should keep it in german for now so everything is consistent
2) While we have a focus on accessible events, those are not the only ones we have

considering those points maybe we should go with "Events in Österreich und Informationen zu deren Barrierefreiheit"?
it does not sounds as snappy as i would like but currently i have no better suggestion :D

mahdikhashan: sure, i agree.

mahdikhashan: image

kadhonn (approved):

kadhonn (approved):

Issue #292, opened 2024-02-07
Collaborate with the other USCC Fellows


Yolgie:
To make the most of your experience in the project and to have the most impact it will be necessary to collaborate with the other fellows of the UpdateSocial Community Call

mahdikhashan:
We, Me an Roba, met after the first lecture of the assigned course. we talked about UX/UI and how we can communicate and collaborate effectively during the program. I shared the progress of the collaboration with integriert  studieren and the possibility of joining together for a meeting. 

We also shared knowledge about UX personas, project sides in general (collectors and publishers) then briefly about HtmlPublisher and the importance of having accessible publisher(s).

We also talked around testing in general, automated tests and a simple user story ("mehr laden" button). We also tried Google Chrome Screen reader on our main website. 

We skimmed over the Frontend and UX issues which we need to collaborate on.

We decided to meet again, I need to share a few links and topics I introduced and follow up with her to make sure she has access to Github project and the board.

mahdikhashan:
I had a short conversation with "JKUCommunity" website and they had already heard about the project and were happy to collaborate. I asked for a member discord's username and I'm still waiting for his acceptance of my request.

kadhonn:
that sounds awesome!
with the jku and that you are already communicating about the next steps :)

mahdikhashan:
at yesterday's second update social lecture, which centered on design processes and solution/problem matching, we (all fellows) engaged in team-building activities. first, we gathered to challenge ourselves with "Marshmallow Challenge". 

It had great learning outcome for me. I discovered that someone in a team should put on the moderator hat to facilitate the process effectively. The process of building the product/service would be optimized by him/her. The goal(s) of the product, wether focused on delivering value or addressing a problem, should be refined with each iteration. we build and rebuild the mvp for the problem. 

If I want to generalise this method to our project, the feature we are currently working on, "event detail element" has a clear  goal. we aim to present as much of the information that we have and is possible (here filterable properties) to the user through the HtmlPublisher. Other variables are A11Y and UX. we want this new feature to be usable and effective to a diverse range of users (regarding our Personas).  

After that, we worked on designing a summer festival to celebrate our achievements with other fellows, I also learned idea generation and prioritising desires to match our goal. we decided whom we want to invite to this event, from our side I took the chance to invite a key member from the LinzStadt, where the majority of our events are sourced from "linztermine.at" this party. I anticipate this will foster a deeper connection, if one does not already exists.

[...]

PR #434, opened 2024-05-19
add 0xa collector

mahdikhashan:
add 0xa collector, addresses #292


kadhonn (changes requested):
looks good in general but you also have to add it to the actual Coordinator so that it will be used!

kadhonn (approved):

PR #464, opened 2024-07-31
add multiplatform docker build and support for podman as container bu…

twatzl:
support build for arm64
support using podman instead of docker (necessary e.g. for using fedora silverblue as dev machine)

### DIFF
--- a/buildSrc/src/main/kotlin/boudicca-springboot-app.gradle.kts
+++ b/buildSrc/src/main/kotlin/boudicca-springboot-app.gradle.kts
@@ -18,10 +18,28 @@ dependencies {
     testImplementation(versionCatalog.findLibrary("spring-boot-starter-test").get())
 }
 
+val containerEngine: String by rootProject.extra
+
 task<Exec>("imageBuild") {
     inputs.file("src/main/docker/Dockerfile")
     inputs.files(tasks.named("bootJar")) //TODO extract to own plugin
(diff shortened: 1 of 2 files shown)


buildSrc/src/main/kotlin/boudicca-springboot-app.gradle.kts
```diff
@@ -18,10 +18,28 @@ dependencies {
...
+
 task<Exec>("imageBuild") {
     inputs.file("src/main/docker/Dockerfile")
     inputs.files(tasks.named("bootJar")) //TODO extract to own plugin
     dependsOn(tasks.named("assemble"))
-    commandLine("docker", "build", "-t", "localhost/boudicca-${project.name}", "-f", "src/main/docker/Dockerfile", ".")
+    commandLine(containerEngine, "build", "-t", "localhost/boudicca-${project.name}", "-f", "src/main/docker/Dockerfile", ".")
 }
 
+task<Exec>("imageBuildMultiplatform") {
```
kadhonn, line 30: mabye we should put into the name the platform actually built? and not only "multiplatform"

kadhonn, line 30: please also add this image build task to the eventcollectors, they sadly still have their own task because they are not a spring boot app (till now at least, something to consider in the future...)

PR #538, opened 2024-10-25
fix cafe traxlmayr lesungen

belehla:


boudicca.events/eventcollectors/src/main/kotlin/events/boudicca/eventcollector/collectors/CafeTraxlmayrCollector.kt
```diff
@@ -74,36 +74,57 @@ class CafeTraxlmayrCollector : TwoStepEventCollector<Element>("cafetraxlmayr") {
...
-                    val name = dateTimeSplit[1]
-                    builder
-                        .copy()
-                        .withName(name)
-                        .withStartDate(startDate)
-                        .build()
+                    null
                 }
+                events.add(builder
+                    .withName(name)
```
kadhonn, line 102: if you resuse a eventbuilder please make a copy before doing so with .copy()! otherwise data from one event could leak to another event, when the second event does not overwrite it

kadhonn (changes requested):
thanks for fixing this!

boudicca.events/eventcollectors/src/main/kotlin/events/boudicca/eventcollector/collectors/CafeTraxlmayrCollector.kt
belehla: thanks for the info, i added a copy() to the builder!

kadhonn (approved):

PR #560, opened 2024-11-29
chore: move versions to version catalogue

Yolgie:


kadhonn (approved):

kadhonn:
> 2 places left where i could not get it to work, do you want to give it a try @kadhonn ?

why did it not work? Oo
those seemed like normal dependencies?

kadhonn:
> 2 places left where i could not get it to work, do you want to give it a try @kadhonn ?

not sure what was wrong with the google dependencies, but for the playwright stuff you just got confused and copy&pasted wrong values

Yolgie:
thank you for fixing it, was too tired and did not see the obvious mistakes ...

PR #604, opened 2025-02-04
create some base tests for eventdb and search service so that basic f…

kadhonn:
…lows are tested


twatzl (approved):
have a look at mockk https://mockk.io/
https://github.com/willowtreeapps/assertk

Yolgie (approved):

PR #608, opened 2025-02-04
fix: fix gradle 9 warnings

Yolgie:


boudicca.base/search-openapi/build.gradle.kts
```diff
@@ -4,4 +4,9 @@ plugins {
 
 dependencies {
     openapi(project(":boudicca.base:search-api"))
-}
\ No newline at end of file
+    testImplementation(libs.junit.jupiter)
+}
```
kadhonn, line 8: but there aren't even tests in this project?

boudicca.base/eventdb/build.gradle.kts
```diff
@@ -7,5 +7,6 @@ dependencies {
     implementation("org.springframework.boot:spring-boot-starter-security")
     implementation(project(":boudicca.base:common-model"))
     implementation(project(":boudicca.base:eventdb-api"))
+    implementation(libs.logback)
```
kadhonn, line 10: that would probably be a good dependency to put into our own gradle plugins?

Yolgie: true, will move it

Issue #370, opened 2024-04-04
split lat lon field into two separate fields


twatzl:
currently we have location.coordinates where lat and lon are combined

we should split this field up to make it easier in publishers to use the data without having to split up lists

afterwards we should have
location.lat
location.lon

Acceptance criteria:
- old field is deprecated
- eventcollectors are updated
- publisher is updated to use the new field

Issue #599, opened 2025-02-04
Add KotlinLogging to replace the LoggerFactory with a more ideomatic approach


Yolgie:
https://github.com/oshai/kotlin-logging

PR #620, opened 2025-02-06
feat(enricher): add osm location enricher (#612)

### LINKED ISSUE #612: Add OSM Location Enricher

Yolgie:
also:
- add multiple SemanticKeys
- add source as VariantConstant
- add some Key, KeyBuilder & StructuredEventBuilder Tests
- make StructuredEventBuilder handle variants better
- add assertk for better assertions
- extract user-agent to Constants.kt
- fix how ListFormat handles null
- update Key constructor to have variants optional
- update StructuredEvent constructor to have data optional
- add encodeUrl to UrlUtils
- convert LocationEnricherUpdater to functional interface
- fix detekt warnings
- extract the @order annotation values to EnricherOrderConstants

Closes #612 and #370


boudicca.base/enricher/src/main/kotlin/base/boudicca/enricher/service/osm/OsmEnricher.kt
```diff
@@ -0,0 +1,193 @@
...
+
+        val response = HttpClient.newHttpClient().send(request, HttpResponse.BodyHandlers.ofString()).body()
+        val json = jacksonObjectMapper().readTree(response) as ArrayNode
+
+        return if (json.isEmpty) null
+        else json.first()["osm_id"].asText()
+    }
+
+    private fun buildAddress(tags: JsonNode): String {
+        val street = (tags["addr:street"]?.asText() ?: "") + " " + (tags["addr:housenumber"]?.asText() ?: "")
```
kadhonn, line 170: we should trim this string, otherwise there could be a leading or trailing whitespace

boudicca.base/enricher/src/main/kotlin/base/boudicca/enricher/service/osm/OsmEnricher.kt
```diff
@@ -0,0 +1,193 @@
...
+        return osmId
+    }
+
+    private fun searchOsmId(event: StructuredEvent): String? {
+        val eventName = event
+            .getProperty(SemanticKeys.LOCATION_NAME_PROPERTY)
+            .map(Pair<Key, String>::second)
+            .joinToString(separator = ",") { it.encodeURL() }
+
+        val query = """https://nominatim.openstreetmap.org/search?q=$eventName&format=json&limit=1"""
```
kadhonn, line 155: we cannot merge it like this. the site specifies a heavy "max 1 request per second" policy here: https://operations.osmfoundation.org/policies/nominatim/

we need to honor this

kadhonn: it also may be that we are not allowed to use it at all, they say as unacceptable use: "Scraping of details The details page is there for debugging only and may not be downloaded automatically."

also they say "As a general rule, bulk geocoding of larger amounts of data is not encouraged. If you have regular geocoding tasks, please, look into alternatives below."

boudicca.base/enricher/src/main/kotlin/base/boudicca/enricher/service/osm/OsmEnricher.kt
```diff
@@ -0,0 +1,193 @@
...
+            .header("User-Agent", Constants.USER_AGENT)
+            .build()
+        val response = HttpClient.newHttpClient().send(request, HttpResponse.BodyHandlers.ofString()).body()
+
+        // parse osm data into our properties
+        val builder = event.toBuilder()
+        val json = jacksonObjectMapper().readTree(response)
+        val elements = json["elements"] as ArrayNode
+        val tags = elements.first()["tags"]
+        val center = elements.first()["center"]
```
kadhonn, line 63: we probably need more safety checks here. what if elements.first() throws an exception? also, what if no center element is available?

boudicca.base/enricher/src/main/kotlin/base/boudicca/enricher/service/osm/OsmEnricher.kt
```diff
@@ -0,0 +1,193 @@
...
+            property = SemanticKeys.LOCATION_COORDINATES_LAT_PROPERTY,
+            value = center["lat"].asText(),
+            variants = sourceVariant
+        ).withProperty(
+            property = SemanticKeys.LOCATION_COORDINATES_LON_PROPERTY,
+            value = center["lon"].asText(),
+            variants = sourceVariant
+        ).withProperty(
+            property = SemanticKeys.SOURCES_PROPERTY,
+            value = currentSources + "Source:OSM:" + json["osm3s"]["copyright"].asText()
```
kadhonn, line 77: uhm... how is the operator precedence here? does this add 2 strings to the sources or only one?

boudicca.base/enricher/src/test/kotlin/base/boudicca/enricher/service/location/LocationEnricherTest.kt
```diff
@@ -96,11 +96,7 @@ class LocationEnricherTest {
     }
 
     private fun createTestEnricher(testData: List<LocationData>): LocationEnricher {
-        val locationEnricher = LocationEnricher(object : LocationEnricherUpdater {
-            override fun updateData(): List<LocationData> {
-                return testData
-            }
-        })
+        val locationEnricher = LocationEnricher { testData }
```
kadhonn, line 99: uhh, i love this!

kadhonn (changes requested):
most of my comments are smaller issues, but the usage policy of nominatim is really worrying

boudicca.base/enricher/src/main/kotlin/base/boudicca/enricher/service/osm/OsmEnricher.kt
```diff
@@ -0,0 +1,193 @@
...
+            .joinToString(separator = ",") { it.encodeURL() }
+
+        val query = """https://nominatim.openstreetmap.org/search?q=$eventName&format=json&limit=1"""
+        val request = HttpRequest.newBuilder()
+            .uri(URI.create(query))
+            .GET()
+            .header("User-Agent", Constants.USER_AGENT)
+            .build()
+
+        val response = HttpClient.newHttpClient().send(request, HttpResponse.BodyHandlers.ofString()).body()
```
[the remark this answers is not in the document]
Yolgie: should we create something like the fetcher here as well?

boudicca.base/enricher/src/main/kotlin/base/boudicca/enricher/service/osm/OsmEnricher.kt
```diff
@@ -0,0 +1,193 @@
...
+        val json = jacksonObjectMapper().readTree(response) as ArrayNode
+
+        return if (json.isEmpty) null
+        else json.first()["osm_id"].asText()
```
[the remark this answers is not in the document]
Yolgie: actually, if there is an answer there is always a osm_id (but this is of course by convention, i could still add the failsafe)

boudicca.base/enricher/src/main/kotlin/base/boudicca/enricher/service/osm/OsmEnricher.kt
Yolgie: yeah, it's as you pointed it out, have changed this

boudicca.base/enricher/src/main/kotlin/base/boudicca/enricher/service/osm/OsmEnricher.kt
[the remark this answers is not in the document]
Yolgie: or maybe pull the Fetcher out of the collectors and into a common part?

kadhonn:
this does currently depend on PR #626

boudicca.base/enricher/src/main/kotlin/base/boudicca/enricher/service/osm/OsmEnricher.kt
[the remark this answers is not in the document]
Yolgie: done in #626

kadhonn (approved):

boudicca.base/common-model/src/main/kotlin/base/boudicca/format/ListFormat.kt
```diff
@@ -8,7 +8,10 @@ package base.boudicca.format
  * all methods may throw exceptions on wrong formatted values
  */
 object ListFormat {
-    fun parseFromString(value: String): List<String> {
```
twatzl, line 11: Why does the parameter need to be nullable here? when does it make sense to parse a null value?

boudicca.base/enricher/src/main/kotlin/base/boudicca/enricher/service/osm/OsmEnricher.kt
```diff
@@ -0,0 +1,218 @@
...
+
+private const val NOMINATIM_BASE_URL = "https://nominatim.boudicca.events"
+
+@Service
+@Order(EnricherOrderConstants.OsmEnricherOrder)
+class OsmEnricher(
+    private val fetcher: Fetcher = createDefaultFetcher()
+) : Enricher {
+    val sourceVariant = listOf(Variant(SOURCE_VARIANT_NAME, "OSM"))
+    val logger = LoggerFactory.getLogger(this::class.java)!!
```
twatzl, line 33: why do we need !! here?
also logger can be moved to the companion object

boudicca.base/enricher/src/main/kotlin/base/boudicca/enricher/service/osm/OsmEnricher.kt
```diff
@@ -0,0 +1,218 @@
...
+
+    val mapper = jacksonObjectMapper()
```
twatzl, line 35: would it make sense to move the mapper to companion object as well or do we need a separate instance for each instance of the class?

boudicca.base/enricher/src/main/kotlin/base/boudicca/enricher/service/osm/OsmEnricher.kt
```diff
@@ -0,0 +1,218 @@
...
+
+    override fun enrich(event: StructuredEvent): StructuredEvent {
+        val osmId: String? = getOrSearchOsmId(event)
```
twatzl, line 38: Do we need to explicitely set the type here? can't it be inferred from getOrSearchOsmId?

boudicca.base/enricher/src/main/kotlin/base/boudicca/enricher/service/osm/OsmEnricher.kt
```diff
@@ -0,0 +1,218 @@
...
+            )
+        }
+
+        return builder.build()
+    }
+
+    private fun getOrSearchOsmId(event: StructuredEvent): String? {
+        // TODO maybe think about taking all osmIds and not only the first :)
+        val osmIds = event.getProperty(SemanticKeys.LOCATION_OSM_ID_PROPERTY)
+        val osmId = if (osmIds.isNotEmpty()) {
```
twatzl, line 149: you could get rid of osmId and just return if ;)

boudicca.base/enricher/src/main/kotlin/base/boudicca/enricher/service/osm/OsmEnricher.kt
```diff
@@ -0,0 +1,218 @@
...
+            builder.updatePropertyIfEmpty(
+                event,
+                SemanticKeys.LOCATION_CITY_PROPERTY,
+                address["city"]?.asText()
+            )
+        }
+        if (tags != null) {
+            builder.updatePropertyIfEmpty(
+                event,
+                SemanticKeys.LOCATION_CONTACT_EMAIL_PROPERTY,
```
twatzl, line 114: Would it make sense to put all the boudicca tags values and the osm tags values in a common configuration class? so it is easier to maintain

could be a simple map<String, Property>

Then it could be intuitively written as

(osm Tag) "wikidata" (maps) to SemanticKeys.LOCATION_WIKIDATA_PROPERTY
contacts:website to SemanticKeys.LOCATION_URL_PROPERTY
...

boudicca.base/enricher/src/main/kotlin/base/boudicca/enricher/service/osm/OsmEnricher.kt
```diff
@@ -0,0 +1,218 @@
...
+        val locationQuery = locationNames
+            .joinToString(separator = ",")
+            .encodeURL()
+
+        val query = """$NOMINATIM_BASE_URL/search?accept-language=de,en&format=jsonv2&limit=1&q=$locationQuery"""
+        val response = fetcher.fetchUrl(query)
+
+        val json = mapper.readTree(response) as ArrayNode
+
+        return if (json.isEmpty) null
```
twatzl, line 174: parentheses are inconsistent within the method :)

boudicca.base/enricher/src/main/kotlin/base/boudicca/enricher/service/osm/OsmEnricher.kt
```diff
@@ -0,0 +1,218 @@
...
+        else {
+            val elem = json.first()
+            val type = elem["osm_type"]?.asText()
+            val id = elem["osm_id"]?.asText()
```
twatzl, line 178: since were using it again here, should we define application wide constants?

boudicca.base/enricher/src/main/kotlin/base/boudicca/enricher/service/osm/OsmEnricher.kt
```diff
@@ -0,0 +1,218 @@
...
+                type[0] + id
+            } else {
+                null
+            }
+        }
+    }
+
+    private fun buildAddress(tags: JsonNode): String {
+        val street = ((tags["road"]?.asText() ?: "") + " " + (tags["house_number"]?.asText() ?: "")).trim()
+        val postcode = tags["postcode"]?.asText()
```
twatzl, line 189: constants?

deployment/DEPLOY.md
```diff
@@ -1,3 +1,27 @@
 # Deployment
 
-TODO add docu
\ No newline at end of file
+TODO add docu
```
twatzl, line 3: is there a ticket for the todo ? :)

(i know it's unrelated to this MR, but would be cool if we could document that eventually)

boudicca.base/enricher/src/main/kotlin/base/boudicca/enricher/service/osm/OsmEnricher.kt
Yolgie: because the LoggerFactory is returning a Java Logger! element and not nullsafe. 

a nicer alternative could be a `requireNotNull` around it 
```suggestion
    val logger = requireNotNull(LoggerFactory.getLogger(this::class.java))
```

boudicca.base/enricher/src/main/kotlin/base/boudicca/enricher/service/osm/OsmEnricher.kt
Yolgie: logger and mapper could be moved to the companion object, the sourceVariant could be moved to the top level declarations (next to NOMINATIM_BASE_URL) or put inline since it only has one use remaining.

boudicca.base/common-model/src/main/kotlin/base/boudicca/model/structured/StructuredEvent.kt
```diff
@@ -134,12 +134,8 @@ data class StructuredEvent(val name: String, val startDate: OffsetDateTime, val
...
         override fun build(): StructuredEvent {
-            if (name == null) {
-                throw IllegalStateException("name cannot be null for an event!")
-            }
-            if (startDate == null) {
-                throw IllegalStateException("startDate cannot be null for an event!")
-            }
+            checkNotNull(name) { "name cannot be null for an event!" }
+            checkNotNull(startDate) { "startDate cannot be null for an event!" }
             return StructuredEvent(name!!, startDate!!, data.toMap())
```
Yolgie, line 139: could also be written as:
```
        override fun build() = StructuredEvent(
            checkNotNull(name) { "name cannot be null for an event!" },
            checkNotNull(startDate) { "startDate cannot be null for an event!" },
            data.toMap()
        )
```

boudicca.base/enricher/src/main/kotlin/base/boudicca/enricher/service/osm/OsmEnricher.kt
```diff
@@ -0,0 +1,218 @@
...
+        }
+
+        // parse osm data into our properties
+        val builder = event.toBuilder()
+        val json = mapper.readTree(response)
+        if (json.isEmpty) {
+            logger.warn("nominatim did not return any results for osm id $osmId")
+            return event
+        }
+        val firstElement = json.first()
```
Yolgie, line 64: would `rootElement` or `nominatimLocationJson` or `locationJson` or `location` or something maybe be better names than firstElement? i'm guessing the first is because there might be multiple results? should there be multiple results, should we use single() instead of first()?

boudicca.base/enricher/src/main/kotlin/base/boudicca/enricher/service/osm/OsmEnricher.kt
```diff
@@ -0,0 +1,218 @@
...
+            .encodeURL()
+
+        val query = """$NOMINATIM_BASE_URL/search?accept-language=de,en&format=jsonv2&limit=1&q=$locationQuery"""
+        val response = fetcher.fetchUrl(query)
+
+        val json = mapper.readTree(response) as ArrayNode
+
+        return if (json.isEmpty) null
+        else {
+            val elem = json.first()
```
Yolgie, line 176: same here `location` would make it more readable than `elem`

boudicca.base/enricher/src/main/kotlin/base/boudicca/enricher/service/osm/OsmEnricher.kt
```diff
@@ -0,0 +1,218 @@
...
+            val type = elem["osm_type"]?.asText()
+            val id = elem["osm_id"]?.asText()
+            if (!type.isNullOrEmpty() && !id.isNullOrEmpty()) {
```
Yolgie, line 179: could they be blank? would `isNullOrBlank` be safer?

Yolgie:
especially with the squashes, feel free to add `Co-authored-by: Matthias 'Yolgie' Holzinger <yolgie@cnoize.at>` to the comment to mark us both as authors :)

Yolgie:
i really love that we have our own nominatim instance to query against ❤️

boudicca.base/enricher/src/main/kotlin/base/boudicca/enricher/service/osm/OsmEnricher.kt
kadhonn: sadly nominatim has it's own data format which changes and keeps osm tags, depending on what the tag is. as you can see we take properties from different json objects here, so a common map does not really work or make it clearer/more reusable

boudicca.base/enricher/src/main/kotlin/base/boudicca/enricher/service/osm/OsmEnricher.kt
kadhonn: i will move it to the companion object, but i will not change the !! for now since we will switch to a better kotlin logger soon(ish) with #599 which should solve this issue

boudicca.base/enricher/src/main/kotlin/base/boudicca/enricher/service/osm/OsmEnricher.kt
kadhonn: i renamed it to `nominatimPlace` because the structure is a "place" as described in https://nominatim.org/release-docs/develop/api/Output/

also since we lookup a single osmId there can only be a single result (or none), so i will switch it to `single()`

boudicca.base/enricher/src/main/kotlin/base/boudicca/enricher/service/osm/OsmEnricher.kt
kadhonn: i would also not use constants for things which are only used in one place and are not completely arbitrary like magic numbers

deployment/DEPLOY.md
kadhonn: we have #282

boudicca.base/enricher/src/main/kotlin/base/boudicca/enricher/service/osm/OsmEnricher.kt
twatzl: ahhh. ok. thats why it looked unfamiliar. i am used to kotlinLogging.

boudicca.base/enricher/src/main/kotlin/base/boudicca/enricher/service/osm/OsmEnricher.kt
twatzl: hmm. what if i want to query other tags e.g. when using museumsbahn-events? is there a generic mechanism for achieving this already? probably not, right?

Yolgie:
can't really review my own pr, so consider this as 👍 approved

boudicca.base/enricher/src/main/kotlin/base/boudicca/enricher/service/osm/OsmEnricher.kt
kadhonn: no generic mechanism for now, no

PR #623, opened 2025-02-06
fix(detekt): add newline to the end of all kt files

Yolgie:
also:
 - add editorconfig for this
 - optimize imports


kadhonn:
uhm.. either it is because i hadn't coffee yet but i think something went wrong? the way i see the diff you REMOVED all newlines at the end?

PR #625, opened 2025-02-06
add new validators to work with variants and properties

### LINKED ISSUE #616: Fix event validation for local debugger
with the introduction of variants the validator broke. 

https://github.com/boudicca-events/boudicca.events/blob/main/boudicca.base/eventcollector-client/src/main/kotlin/base/boudicca/api/eventcollector/debugger/EventCollectorValidation.kt

for usage example see here

https://github.com/museumsbahn-it-community/museumsbahn-events/blob/main/museum-railway-eventcollectors/src/main/kotlin/at/museumrailwayevents/eventcollectors/LocalCollectorDebug.kt

fix the validations, so they support variants again

kadhonn:
resolves #616 by adding some new validators to work with the new properties/variants

### DIFF
--- a/boudicca.base/common-model/src/main/kotlin/base/boudicca/model/structured/Key.kt
+++ b/boudicca.base/common-model/src/main/kotlin/base/boudicca/model/structured/Key.kt
@@ -63,4 +63,6 @@ data class Key(val name: String, val variants: List<Variant>) : Comparable<Key> 
             return Key(propertyName, variants.toList().sorted())
         }
     }
+
+    override fun toString() = toKeyString()
 }
--- a/boudicca.base/eventcollector-client/src/main/kotlin/base/boudicca/api/eventcollector/debugger/DataShouldContainKey.kt
+++ b/boudicca.base/eventcollector-client/src/main/kotlin/base/boudicca/api/eventcollector/debugger/DataShouldContainKey.kt
@@ -6,9 +6,9 @@ import base.boudicca.api.eventcollector.debugger.color.yellow
 import base.boudicca.model.Event
 
 class DataShouldContainKey(
-    val key: String,
-    val severity: ValidationSeverity,
-): EventCollectorValidation {
+    private val key: String,
+    private val severity: ValidationSeverity,
+) : EventCollectorValidation {
     override fun validate(event: Event, verbose: Boolean): ValidationResult {
         if (!event.data.containsKey(key)) {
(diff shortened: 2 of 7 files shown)


Yolgie:
new files should end with new line, the editorconfig that has been checked in will auto do this on format file in the future, but has not been merged yet :)

boudicca.base/common-model/src/main/kotlin/base/boudicca/model/structured/Key.kt
```diff
@@ -63,4 +63,6 @@ data class Key(val name: String, val variants: List<Variant>) : Comparable<Key>
             return Key(propertyName, variants.toList().sorted())
         }
     }
+
+    override fun toString() = toKeyString()
```
Yolgie, line 67: what do we need the `toKeyString()` for, would `toString()` not be enough?

Yolgie:
- `EventCollectorValidation` can be a functional interface 
- verbose should have a default value imho

boudicca.base/common-model/src/main/kotlin/base/boudicca/model/structured/Key.kt
kadhonn: toString() normally is for debug-logging and not for serializing stuff into whatever technical format, which is why i feel uncomfortable using toString() instead of toKeyString() when i want to make sure it corresponds to the right format. here it is coincidence that toString() does the same as toKeyString(), which is why i delegate it

kadhonn:
> * `EventCollectorValidation` can be a functional interface
yup, done

>     * verbose should have a default value imho
you cannot do that for functional interfaces (which it is now)

Yolgie:
> > * `EventCollectorValidation` can be a functional interface
> yup, done
> 
> >     * verbose should have a default value imho
> you cannot do that for functional interfaces (which it is now)

But the implementing classes can still do that 😊

kadhonn:
> > > * `EventCollectorValidation` can be a functional interface
> > >   yup, done
> > 
> > 
> > > ```
> > > * verbose should have a default value imho
> > > ```
> > 
> > 
> > you cannot do that for functional interfaces (which it is now)
> 
> But the implementing classes can still do that 😊

which does not matter because noone will use the overloaded method ever? but only the interface method where you have to specify it?

Yolgie (approved):

PR #634, opened 2025-02-20
add first version of a map-view of events with leaflet

kadhonn:
also some small changes to the osmenricher and some collectors


boudicca.base/publisher-event-html/src/main/kotlin/base/boudicca/publisher/event/html/model/MapSearchResultDTO.kt
```diff
@@ -0,0 +1,20 @@
+package base.boudicca.publisher.event.html.model
+
+
+data class MapSearchResultDTO(
```
Yolgie, line 4: while we use this for a map, would `LocationSearchResult` and `/locationSearch` not be better namings for the DTO and endpoint?

boudicca.base/publisher-event-html/src/main/kotlin/base/boudicca/publisher/event/html/service/EventService.kt
```diff
@@ -44,25 +48,33 @@ class EventService @Autowired constructor(
...
-        return mapEvents(events)
+        val searchResult = caller.search(QueryDTO(generateQuery(searchDTO), searchDTO.offset ?: 0))
+        return mapEvents(searchResult)
     }
 
-    fun generateQuery(searchDTO: SearchDTO): String {
+    @Throws(EventServiceException::class)
+    fun mapSearch(searchDTO: SearchDTO): MapSearchResultDTO {
+        val searchResultDTO =
+            caller.search(QueryDTO(generateQuery(searchDTO, additionalMapFilters), searchDTO.offset ?: 0, 200))
```
Yolgie, line 62: what does the 200 mean? flagged as magic-number :)

boudicca.base/publisher-event-html/src/main/kotlin/base/boudicca/publisher/event/html/service/EventService.kt
```diff
@@ -44,25 +48,33 @@ class EventService @Autowired constructor(
...
+        return mapMapSearch(searchResultDTO)
+    }
+
+    fun generateQuery(searchDTO: SearchDTO, additionalParts: List<String> = emptyList()): String {
```
Yolgie, line 66: can we find a better name than additionalParts? i have no clue what this should mean

Yolgie (changes requested):
only looked at the .kt parts, brain to fuzzy to look at js today...

boudicca.base/publisher-event-html/src/main/kotlin/base/boudicca/publisher/event/html/service/EventService.kt
kadhonn: well, it is a magic number, namely the max results we wanna show :D i did a lot here since only 30 on the whole map is not enough. i will add a constant

boudicca.base/publisher-event-html/src/main/kotlin/base/boudicca/publisher/event/html/model/MapSearchResultDTO.kt
kadhonn: hm... good point, but i think `locationSearch` is a bit misleading because you don't actually search for locations, but the result is grouped by locations... but yeah, MapSearch is also not a great name... will think a bit about that one :D

kadhonn: hm... i think i would leave it as `mapSearch` for now. the reason is that this is only an internal and not a public api (this is in the html-publisher) so it will only ever be used for the map and can be quite specific to that

boudicca.base/publisher-event-html/src/main/resources/templates/layout/map.hbs
```diff
@@ -0,0 +1,19 @@
+<style>
+    #map {
+        height: 600px;
+    }
+    .leaflet-popup-content {
+        ul {
```
belehla, line 6: css nesting doesn't work for me in firefox: 
image
(but it works in chrome)

docs/QUERY.md
```diff
@@ -56,6 +56,6 @@ You can use the grouping `(...)` mechanism to circumvent the order.
 * Search for any event while I am in Vienna on
   holiday: `"location.city" equals "Wien" and "startDate" after "2023-05-27" and "startDate" before "2023-05-31"`
 * Search for events with a duration of 2 hours or less: `duration "startDate" "endDate" shorter 2`
-* Search for events which have a english description: `hasProperty "description:lang=en"`
+* Search for events which have a english description: `hasField "description:lang=en"`
```
belehla, line 59: i think it should be: "an English description"

boudicca.base/publisher-event-html/src/main/resources/templates/layout/map.hbs
kadhonn: oh really? it works in my firefox oO
what version of firefox are you using?

belehla: i used version 115
i just updated to 128 and now it works.. maybe we still change it to .leaflet-popup-content ul{} to support older versions

kadhonn: yeah, sad but we should change it, will do so tomorrow

Yolgie (approved):

PR #639, opened 2025-02-24
added description, url and location to ical events

### LINKED ISSUE #618: ical-publisher: Add description and link to event page, allow custom ical title
Right now the events in the ical published calendar are rather useless.

They have neither a description nor a link to where to find more information.

Add an additional field publisher.ical.displayname for setting a custom name for ical publisher.
Display a link to the original source and add the possiblity to add an link to the published page. maybe as publisher.ical.moreInfoUrl.

Image

kadhonn:
closes #618

### DIFF
--- a/boudicca.base/publisher-event-ical/src/main/kotlin/base/boudicca/publisher/event/ical/CalendarService.kt
+++ b/boudicca.base/publisher-event-ical/src/main/kotlin/base/boudicca/publisher/event/ical/CalendarService.kt
@@ -4,21 +4,19 @@ import base.boudicca.SemanticKeys
 import base.boudicca.api.search.QueryDTO
 import base.boudicca.api.search.SearchClient
 import base.boudicca.model.Event
+import base.boudicca.model.structured.Key
+import base.boudicca.model.structured.StructuredEvent
 import biweekly.Biweekly
 import biweekly.ICalVersion
 import biweekly.ICalendar
 import biweekly.component.VEvent
-import biweekly.property.DateEnd
-import biweekly.property.DateStart
-import biweekly.property.Location
-import biweekly.property.Sequence
-import biweekly.property.Uid
+import biweekly.property.*
 import org.springframework.beans.factory.annotation.Autowired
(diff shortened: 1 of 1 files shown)


twatzl:
do you think it would make sense to define properties like

ical.name, ical.description and maybe ical.url to allow "manually" overriding the fields e.g. by an enricher?

The reason for that is there might be multiple events in multiple locations with similiar names and then it might make sense to add some additional information to the name.

image

And yes there is the location name, but again that is difficult because the location name might be completely different than the operator. :D

Hard to explain. Maybe have a short call?

boudicca.base/publisher-event-ical/src/main/kotlin/base/boudicca/publisher/event/ical/CalendarService.kt
```diff
@@ -32,56 +30,56 @@ class CalendarService @Autowired constructor(@Value("\${boudicca.search.url}") p
...
     }
 
     fun getEvents(query: String): ByteArray {
-        val events = searchClient.queryEvents(QueryDTO(query,0, Int.MAX_VALUE))
+        val events = searchClient.queryEvents(QueryDTO(query, 0, Int.MAX_VALUE))
         return createCalendar(events.result)
     }
 }
+
+private fun OffsetDateTime.toDate() = Date.from(this.toInstant())
```
Yolgie, line 85: this could go into a helper

Yolgie (approved):

boudicca.base/publisher-event-ical/src/main/kotlin/base/boudicca/publisher/event/ical/CalendarService.kt
kadhonn: not sure, it's just a small line and there is no other class in the ical-project. not sure if this is important enough to put into the common-model, which is the only other place we could put it

kadhonn:
@twatzl i will merge this but will keep the story open because your problem is not solved yet

PR #649, opened 2025-02-26
Abl/springboot lib

kadhonn:
well, i wanted to do something completely different but i found more and more issues i had to solve until i can start there :D 

so this PR is a lot of refactoring, it probably makes more sense to look at the individual commits and not at the whole diff

the main changes are:
1) i split the "imageBuild" task for docker builds out into it's own plugin with their configuration. that means your multiplatform docker build is now easier to maintain and create @twatzl 
2) i fixed a small issue which broke the "gradle clean build" execution, that should now work again
3) the biggest change was to the html-publisher, i split it into some sort of "publisher library" and the actual "publisher app", which makes it way easier to extend it and have multiple versions. in the library there is pretty much all the code + a spring file which automatically loads the main configuration if the library is on the classpath. so the "app" part only needs a main class and a dependency to the "lib"

and some other smaller cleanups if i happen to see something to cleanup


boudicca.base/publisher-event-html-app/.gitignore
```diff
@@ -0,0 +1,37 @@
+HELP.md
```
Yolgie, line 1: why does the publisher have it's own .gitignore?

kadhonn: i have no idea :D i removed all unneeded .gitignore files, there where more of them

PR #656, opened 2025-03-01
split up enricher into lib and boudicca.events enricher

kadhonn:
i did split the enricher into a library which is just an empty controller + service, but has all the apis needed to build an enricher yourself.

the other part are our actual enrichers, which i have moved to the boudicca.events project, because i think those should not be hosted by everyone themselves, but only once by us. if someone wants to use those they can simply use our enricher. at least this is the case for the more involved ones like the location/osm/musicbrainz enrichers. 

what do you think?


twatzl:
looking at EnricherService.. is the @Autowired there really necessary? isn't that taken care of by @Service annotation already? also constructor keyword there is unnecessary.

instead of
```kotlin
@Service
class EnricherService @Autowired constructor(
    private val enrichers: List<Enricher>,
    private val eventPublisher: ApplicationEventPublisher,
) {
```

it should be sufficient to write
```kotlin
@Service
class EnricherService(
    private val enrichers: List<Enricher>,
    private val eventPublisher: ApplicationEventPublisher,
) {
```

boudicca.base/enricher/src/main/kotlin/base/boudicca/enricher/EnricherConfiguration.kt
```diff
@@ -4,8 +4,11 @@ import io.swagger.v3.oas.annotations.OpenAPIDefinition
...
 import io.swagger.v3.oas.annotations.info.License
 import io.swagger.v3.oas.annotations.servers.Server
+import org.springframework.boot.autoconfigure.AutoConfiguration
 import org.springframework.boot.autoconfigure.SpringBootApplication
 import org.springframework.boot.runApplication
+import org.springframework.context.annotation.ComponentScan
+import org.springframework.context.annotation.Configuration
 import org.springframework.scheduling.annotation.EnableScheduling
 
 @OpenAPIDefinition(
```
twatzl, line 14: Why is the OpenApiDefinition part of the configuration? Shouldn't that be in the enricher-api module provided as interface and then used in the controller?

boudicca.base/enricher/src/main/kotlin/base/boudicca/enricher/EnricherConfiguration.kt
```diff
@@ -20,10 +23,6 @@ import org.springframework.scheduling.annotation.EnableScheduling
...
 )
-@SpringBootApplication
-@EnableScheduling
-class EnricherApplication
-
-fun main(args: Array<String>) {
-    runApplication<EnricherApplication>(*args)
-}
+@AutoConfiguration
+@ComponentScan
```
twatzl, line 27: Should we really use component scan here? to me it seems this is obfuscating a lot of magic that happens under the hood.

When implementing an enricher for some kind of boudicca derivative service I would like to choose which enrichers I want to use and not magically pull in all of them, right?

twatzl:
I am missing documentation on how the different modules work together here. We have some high level explanations on how the enrichers work, but nothing specific. It would be cool to have a code example how to implement a basic custom enricher and how to bundle it together with the other ones. Maybe also explain the modules on a basic level
we have

```
/boudicca
    /enricher
    /enricher-api
    /enricher-client
    /enricher-openapi
    /enricher-utils
/boudicca.events
    /enricher
    /enricher-utils
```

what goes where?
how to generate new openapi?
how to update client?

boudicca.events/enricher/src/main/kotlin/events/boudicca/enricher/service/osm/OsmEnricher.kt
```diff
@@ -17,6 +16,7 @@ import base.boudicca.model.structured.VariantConstants.SOURCE_VARIANT_NAME
 import com.fasterxml.jackson.databind.JsonNode
 import com.fasterxml.jackson.databind.node.ArrayNode
 import com.fasterxml.jackson.module.kotlin.jacksonObjectMapper
+import events.boudicca.enricher.service.EnricherOrderConstants
 import io.github.oshai.kotlinlogging.KotlinLogging
 import org.springframework.core.annotation.Order
 import org.springframework.stereotype.Service
```
twatzl, line 22: why is the OSMEnricher not part of the lib anymore, but now boudicca.events specific? IMO it could be reused by other boudicca applications, not?

boudicca.events/enricher/.dockerignore
```diff
@@ -1,2 +1,2 @@
-*
```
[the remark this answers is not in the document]
kadhonn: wtf, i did not make that change, intellij must have done some automagic... thanks for noticing!

buildSrc/src/main/kotlin/boudicca-springboot-rest-lib.gradle.kts
```diff
@@ -0,0 +1,14 @@
...
+ * plugin for springboot "libraries" with an openapi rest api
+ */
+
+plugins {
+    id("boudicca-springboot-lib")
+}
+
+val versionCatalog = versionCatalogs.named("libs")
+dependencies {
+    api(versionCatalog.findLibrary("spring-boot-starter-web").get())
```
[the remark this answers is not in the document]
kadhonn: because that is the only way it works when you are applying it in a script-plugin

boudicca.base/enricher/src/main/kotlin/base/boudicca/enricher/EnricherConfiguration.kt
kadhonn: i cannot remember exactly why i did it this way but i only know that other ways did not work... maybe they work now but i am not motivated to spend more time on this

boudicca.base/enricher/src/main/kotlin/base/boudicca/enricher/EnricherConfiguration.kt
kadhonn: "When implementing an enricher for some kind of boudicca derivative service I would like to choose which enrichers I want to use and not magically pull in all of them, right?"

which is exactly why i made the split. the `ComponentScan` only detects the controller and the enricherservice itself, not actual implementations of enrichers, which now are part of the actual enricher, which is custom to boudicca.events and whoever wants to do their own implementation

boudicca.events/enricher/src/main/kotlin/events/boudicca/enricher/service/osm/OsmEnricher.kt
kadhonn: because i don't think it makes sense to implement enrichers more then once, especially if a lot of infrastructure is needed for that. for example for the OSM Enricher you need your own Nominatim instance, which is something i setup for us but others probably won't like to do. they should just use the osm enricher we are using

kadhonn:
> I am missing documentation on how the different modules work together here. We have some high level explanations on how the enrichers work, but nothing specific. It would be cool to have a code example how to implement a basic custom enricher and how to bundle it together with the other ones. Maybe also explain the modules on a basic level we have
> 
> ```
> /boudicca
>     /enricher
>     /enricher-api
>     /enricher-client
>     /enricher-openapi
>     /enricher-utils
> /boudicca.events
>     /enricher
>     /enricher-utils
> ```

well, those are the same modules we have everywhere, -api is the interfaces, -openapi is the generated openapi schema and -client is a "nice" custom written client for the enricher, same as with search/eventdb
the only unusual repo is "utils" which are some utils needed for importing musicbrainz data needed for our ArtistEnricher, but this is simply a static class

also, 
> /boudicca.events
>     /enricher
> 
would be your basic example how to implement your own enricher, but yes we could add a better example for this. but i don't see this part of this story

> what goes where? how to generate new openapi? how to update client?

no idea what you mean here?

PR #657, opened 2025-03-01
implement dsl for creating structured events

twatzl:
this is a draft and request for comments.

I have drafted how a kotlin DSL for creating structured events could look like.


boudicca.base/common-model/src/main/kotlin/base/boudicca/model/structured/StructuredEventBuilder.kt
```diff
@@ -0,0 +1,26 @@
...
+) : AbstractStructuredBuilder<StructuredEvent, StructuredEventBuilder>(data.toMutableMap()) {
+
+    fun withData(name: String, init: DataBuilder.() -> Unit) {
+        val builder = DataBuilder(name)
+        builder.init()
+        val dataEntries = builder.build()
+
+        this.data.putAll(dataEntries)
+    }
+
```
kadhonn, line 18: why did you remove the copy()/withName()/withStartDate() methods?

boudicca.base/common-model/src/main/kotlin/base/boudicca/model/structured/VariantConstants.kt
```diff
@@ -17,6 +17,8 @@ object VariantConstants {
         const val NUMBER_FORMAT_NAME = "number"
         const val DATE_FORMAT_NAME = "date"
         const val LIST_FORMAT_NAME = "list"
+        const val ENUM_FORMAT_NAME = "enum"
```
kadhonn, line 20: what does format=enum mean? what enum?

boudicca.base/common-model/src/main/kotlin/base/boudicca/model/structured/VariantUtils.kt
```diff
@@ -0,0 +1,30 @@
...
+fun listFormat(): Variant {
+    return Variant(VariantConstants.FORMAT_VARIANT_NAME, VariantConstants.FormatVariantConstants.MARKDOWN_FORMAT_NAME)
+}
+
+fun jsonFormat(): Variant {
+    return Variant(VariantConstants.FORMAT_VARIANT_NAME, VariantConstants.FormatVariantConstants.MARKDOWN_FORMAT_NAME)
+}
+
+fun enumFormat(): Variant {
+    return Variant(VariantConstants.FORMAT_VARIANT_NAME, VariantConstants.FormatVariantConstants.MARKDOWN_FORMAT_NAME)
```
kadhonn, line 20: enumFormat() and jsonFormat() and listFormat() all set the wrong variant value

boudicca.base/common-model/src/test/kotlin/base/boudicca/model/StructuredEventDSLTest.kt
```diff
@@ -0,0 +1,39 @@
...
+                )
+                variant(
+                    lang("de"),
+                    data = "MEINE DATEN!"
+                )
+            }
+            withData(SemanticKeys.SOURCES) {
+                variant(
+                    listFormat(),
+                    data = "[asdf.com, bsdf.com, csdf.com]"
```
kadhonn, line 30: this is why i really want typed setters for this. this value will not result in the list think it does (pretty sure at least)

boudicca.base/common-model/src/main/kotlin/base/boudicca/model/structured/StructuredEntryDSL.kt
```diff
@@ -0,0 +1,9 @@
+package base.boudicca.model.structured
+
+import java.time.OffsetDateTime
+
+fun structuredEvent(name: String, startDate: OffsetDateTime, init: StructuredEventBuilder.() -> Unit = {}): StructuredEvent {
```
kadhonn, line 5: file is named wrong, event!=entry

kadhonn (commented):
while i really like the dsl in general, i don't think it replaces the "properties" concept because at least as it is now it is way more verbose and errorprone, because you have to set all variants yourself, and not mix anything up. oh, and it does not have type safety currently

i really like it as a builder and as a possible other way for the `withKeyValuePair` method we also currently have

boudicca.base/common-model/src/main/kotlin/base/boudicca/model/structured/StructuredEventBuilder.kt
twatzl: Why do we need them?
Is there really a case where event name and start date are unknown at event creation time?

This way we can get rid of the nullability, which resolves a lot of headaches.

Similarly I do not see the use case for a copy operation?

boudicca.base/common-model/src/main/kotlin/base/boudicca/model/structured/VariantConstants.kt
twatzl: didn't we also discuss enums and json as formats for variants when we originally envisioned these variants?
Mainly we need to find a generic way to enter enums

boudicca.base/common-model/src/main/kotlin/base/boudicca/model/structured/StructuredEntryDSL.kt
twatzl: will be fixed

boudicca.base/common-model/src/test/kotlin/base/boudicca/model/StructuredEventDSLTest.kt
twatzl: will be fixed

twatzl:
pushed an update. it should now be typesafe.

we still should think about enum and json types as i would like to provide a clean and structured interface

boudicca.base/common-model/src/main/kotlin/base/boudicca/model/structured/StructuredEventBuilder.kt
kadhonn: you could just look at the usages of the methods :P in our case collectors like BrucknerhausCollector benefit from that. they have the same event at different starttimes, so they create the full event, and then iterate over all starttimes, copy the builder, set the time and call .build()

boudicca.base/common-model/src/main/kotlin/base/boudicca/model/structured/VariantConstants.kt
kadhonn: yes we did discuss it, and we have some in our semantic conventions as well. but i am not sure what marking as a format brings us. maybe just for metadata purposes, so someone knows somewhere is a enum defined? but i am really not sure

boudicca.base/common-model/src/main/kotlin/base/boudicca/model/structured/StructuredEventBuilder.kt
twatzl: Hmm. I was looking for usages, but couldn't find any. thats why i removed the. maybe i have to look again.

boudicca.base/common-model/src/main/kotlin/base/boudicca/format/DateFormatAdapter.kt
```diff
@@ -11,17 +12,16 @@ import java.time.format.DateTimeParseException
  *
  * all methods may throw exceptions on wrong formatted values
  */
-object DateFormat {
```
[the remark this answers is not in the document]
twatzl: Honestly I don't really see the point of keeping them in objects rather than in classes. All the AbstractFormatAdapter does is force "the user" to conform to a common interface that has to be fulfilled.

There is not more or less logic than before.

The only addition is that now there is a common class which can be used for parameters etc.

Also the usage is not more complicated.

before it was
```
DateFormat.parseFromString()
```

and now it became
```
DateFormatAdapter().parseFromString()
```

Testability also stays the same. The only addition is a common interface that forces developers to add a tiny bit of metadata as well.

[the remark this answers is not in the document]
kadhonn: i still see those as 2 different aspects. the first one is the parsing logic, the second one is the adapter-hierarchy. the main reason for separating these two is that one should not have to change with the other. you had to adapt and change a lot of places where those objects where already used, which you would have not to do if you simply delegated it to the objects. and similarly in the future, if we change something with the adapters we have to change everything which uses the parsing logic, and not only the DSL relevant classes

boudicca.base/common-model/src/main/kotlin/base/boudicca/model/structured/VariantConstants.kt
twatzl: i created a separate ticket for it. added them now for completeness, but left them empty as to indicate use of text format.

twatzl:
> while i really like the dsl in general, i don't think it replaces the "properties" concept because at least as it is now it is way more verbose and errorprone, because you have to set all variants yourself, and not mix anything up. oh, and it does not have type safety currently
> 
> i really like it as a builder and as a possible other way for the `withKeyValuePair` method we also currently have

Is this still relevant? 
As it is right now the "old" approach with properties still works. just the syntax is a tiny bit different and you save 2 lines.

kadhonn (changes requested):

boudicca.base/common-model/src/main/kotlin/base/boudicca/model/structured/StructuredEventBuilder.kt
kadhonn: while the change is fine with me "the new way should also have slightly improved memory performance." is not true since now you parse urls and execute CSS selectors in a loop instead of once :P

kadhonn (approved):

PR #666, opened 2025-03-09
split up Keys and KeyFilters into two seperate classes

kadhonn:


boudicca.base/common-model/src/main/kotlin/base/boudicca/model/structured/KeyFilter.kt
```diff
@@ -0,0 +1,26 @@
...
+        }
+
+        fun builder(propertyName: String): KeyFilterBuilder {
+            return KeyFilterBuilder(propertyName)
+        }
+    }
+
+    override fun toBuilder() = KeyFilterBuilder(name).withVariants(variants)
+
+    class KeyFilterBuilder(name: String) : AbstractKeyBuilder<KeyFilter>(name) {
```
twatzl, line 21: So if we did that the Kotlin way we would get rid of the "toBuilder" function and move the KeyFilterBuilder to a separate file.

Then define a `fun keyFilter(keyFilter: KeyFilter, init KeyFilterBuilder.() -> {})` (or call it modifyKeyFilter if you want and keyFilter for creating new filters)

boudicca.base/common-model/src/main/kotlin/base/boudicca/model/structured/AbstractKey.kt
```diff
@@ -0,0 +1,97 @@
...
+        return true
+    }
+
+    override fun hashCode(): Int {
+        var result = name.hashCode()
+        result = 31 * result + variants.hashCode()
+        return result
+    }
+
+    abstract class AbstractKeyBuilder<T>(private val name: String) {
```
twatzl, line 74: here as well i would recommend not having nested classes, but rather go the kotlin way and move it to a separate file

boudicca.base/common-model/src/main/kotlin/base/boudicca/model/structured/AbstractKey.kt
```diff
@@ -0,0 +1,97 @@
...
+    fun toKeyString(): String {
+        if (variants.isEmpty()) {
+            return name
+        }
+        return name + ":" + variants.joinToString(":") { it.toKeyString() }
+    }
+
+    companion object {
+        val COMPARATOR = compareBy<AbstractKey<*>> { it.name }.thenComparing { o1, o2 ->
+            for (i in 0..<min(o1.variants.size, o2.variants.size)) {
```
twatzl, line 29: you don't need a for loop here as progressions are iterable
```suggestion
           (0..min(o1.variants.size, o2.variants.size)).forEach { i ->
```

boudicca.base/common-model/src/main/kotlin/base/boudicca/model/structured/AbstractKey.kt
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    /**
+     * get the key string representation of this key for using in an unstructured/serialized event
+     */
+    fun toKeyString(): String {
+        if (variants.isEmpty()) {
+            return name
+        }
+        return name + ":" + variants.joinToString(":") { it.toKeyString() }
```
twatzl, line 24:
```suggestion
        return "$name:${variants.joinToString(":") { it.toKeyString() }}"
```

boudicca.base/common-model/src/main/kotlin/base/boudicca/keyfilters/KeySelector.kt
```diff
@@ -105,6 +99,12 @@ class KeySelector private constructor(
             key.variants.forEach { builder.thenVariant(it) }
             return builder
         }
+
+        fun builder(keyFilter: KeyFilter): KeySelectorBuilder {
```
twatzl, line 103: while we're at it we could also extract this builder?

Also is there a reason why i would need to use the builder and cannot just pass a name and list to the constructor?

kadhonn: yup, i will extract it. hm, not really. i think it makes it a bit more easier to reason about in this case, but i will make the constructor public, so you don't have to use the builder

twatzl (approved):

PR #667, opened 2025-03-09
Abl/css layers

kadhonn:


twatzl (approved):
Looks good on a quick glance. Don't have much in depth knowledge of how the boudicca frontend works.

I saw you replaced some ids with classes. This is good. In general we should always use classes for styling. Ids only in very rare circumstances.

kadhonn:
yeah, the problem was that those id's where rendered in a loop, so they were not even unique...^^

PR #671, opened 2025-03-09
draft: WIP Tw/common model js

twatzl:
attempt to convert common model to kotlin multiplatform


twatzl:
Currently the big question is how we want to proceed on this since the change to a common self implemented offsetDateTime would mean a change in the API and be quite a lot of work.

PR #677, opened 2025-03-17
reenable detekt with relaxed ruleset and fix all remaining issues

kadhonn:


boudicca.base/common-model/src/main/kotlin/base/boudicca/model/structured/AbstractKey.kt
```diff
@@ -26,7 +26,7 @@ abstract class AbstractKey<T : AbstractKey<T>>(val name: String, variants: List<
 
     companion object {
         val COMPARATOR = compareBy<AbstractKey<*>> { it.name }.thenComparing { o1, o2 ->
-            (0..<min(o1.variants.size, o2.variants.size)).forEach { i ->
+            for (i in 0..<min(o1.variants.size, o2.variants.size)) {
```
twatzl, line 29: why?

boudicca.base/common-model/src/main/kotlin/base/boudicca/model/structured/StructuredEntry.kt
```diff
@@ -5,13 +5,16 @@ import base.boudicca.keyfilters.KeyFilters
...
 import base.boudicca.model.structured.dsl.StructuredEntryBuilder
+import io.github.oshai.kotlinlogging.KotlinLogging
 import java.util.*
 
 /**
  * as with the [Entry] class, a StructuredEntry is simple a typealias for a Map<Key, String>
  */
 typealias StructuredEntry = Map<Key, String>
 
+private val logger = KotlinLogging.logger("base.boudicca.model.structured.StructuredEntry")
```
twatzl, line 16: do we really need to set the logger name manually? isn't that taken automatically from the classname usually?

boudicca.base/common-model/src/main/kotlin/base/boudicca/model/structured/StructuredEvent.kt
```diff
@@ -45,7 +49,8 @@ data class StructuredEvent(val name: String, val startDate: OffsetDateTime, val
                 try {
                     val parsedValue = property.parseFromString(it.second)
                     Pair(it.first, parsedValue)
-                } catch (_: IllegalArgumentException) {
+                } catch (e: IllegalArgumentException) {
+                    logger.warn(e) { "error parsing value for key '${it.first}': ${it.second}" }
```
twatzl, line 53: should we document with examples what log levels should be used where?

boudicca.base/eventcollector-client/src/main/kotlin/base/boudicca/api/eventcollector/collections/CollectionClasses.kt
```diff
@@ -23,7 +23,8 @@ data class FullCollection(
     )
 
     override fun toString(): String {
-        return "FullCollection(\nid=$id, \nstartTime=$startTime, \nendTime=$endTime, \nsingleCollections=$singleCollections, \nlogLines=$logLines, \nerrorCount=$errorCount, \nwarningCount=$warningCount)"
+        return "FullCollection(\nid=$id, \nstartTime=$startTime, \nendTime=$endTime, \nsingleCollections=$singleCollections" +
```
twatzl, line 26: with so many \n should we rather use a multiline string and trimindent?

boudicca.base/fetcher-lib/src/main/kotlin/base/boudicca/fetcher/Fetcher.kt
```diff
@@ -10,9 +10,13 @@ import java.net.http.HttpResponse.BodyHandlers
 import java.time.Clock
 import java.util.concurrent.Callable
 
+private const val WAIT_TIME_FACTOR = 1.5F
+private const val MIN_WAIT_TIME = 100L
+
 /**
  * THIS CLASS IS NOT THREADSAFE!
  */
+@Suppress("detekt:LongParameterList")
```
twatzl, line 19: instead of supressing, why no create a data class called FetcherConfig?

not sure if detekt also complains about data classes, but maybe this would be a bit cleaner anyway?

boudicca.base/fetcher-lib/src/main/kotlin/base/boudicca/fetcher/Fetcher.kt
```diff
@@ -10,9 +10,13 @@ import java.net.http.HttpResponse.BodyHandlers
 import java.time.Clock
 import java.util.concurrent.Callable
 
+private const val WAIT_TIME_FACTOR = 1.5F
+private const val MIN_WAIT_TIME = 100L
```
twatzl, line 14: is the suggested case of parameters, enum etc. enforced by detekt? https://kotlinlang.org/docs/coding-conventions.html#property-names

boudicca.base/publisher-event-html/src/main/kotlin/base/boudicca/publisher/event/html/service/EventService.kt
```diff
@@ -42,7 +43,9 @@ private const val SEARCH_TYPE_ALL = "ALL"
 
 private const val MAP_SEARCH_RESULT_COUNT = 200
 
+//TODO we should think about reducing the size of this class, maybe we can split out all the property selecting methods
```
twatzl, line 46: pls create ticket

boudicca.base/publisher-event-html/src/main/kotlin/base/boudicca/publisher/event/html/service/EventService.kt
```diff
@@ -83,6 +87,7 @@ class EventService @Autowired constructor(
         return query
     }
 
+    @Suppress("detekt:CyclomaticComplexMethod")
```
twatzl, line 90: we should refactor this function. I too think it could be written better

boudicca.base/query-lib/src/main/kotlin/base/boudicca/query/BoudiccaQueryBuilder.kt
```diff
@@ -45,17 +43,14 @@ object BoudiccaQueryBuilder {
         return escapeText(dateFieldName) + " before " + escapeText(DateTimeFormatter.ISO_LOCAL_DATE.format(localDate))
     }
 
+    @Suppress("detekt:ExceptionRaisedInUnexpectedLocation")
```
twatzl, line 46: why supress when fixed?

boudicca.base/query-lib/src/main/kotlin/base/boudicca/query/evaluator/OptimizingEvaluator.kt
```diff
@@ -30,6 +30,9 @@ import java.time.format.DateTimeParseException
 import java.util.BitSet
 import java.util.concurrent.ConcurrentHashMap
 
+private const val SORTING_RATIO = 3
+
+@Suppress("detekt:TooManyFunctions")
```
twatzl, line 35: do we really need this rule?

boudicca.base/query-lib/src/main/kotlin/base/boudicca/query/evaluator/SimpleEvaluator.kt
```diff
@@ -8,26 +8,15 @@ import base.boudicca.model.structured.StructuredEntry
...
-import base.boudicca.query.Utils
+import base.boudicca.query.*
 import base.boudicca.query.evaluator.util.EvaluatorUtil
 import java.time.LocalDate
 import java.time.OffsetDateTime
 import java.time.ZoneId
 import java.time.format.DateTimeParseException
 import java.util.concurrent.ConcurrentHashMap
 
+@Suppress("detekt:LongMethod", "detekt:CyclomaticComplexMethod")
```
twatzl, line 19: maybe we should also refactor this?

boudicca.base/query-lib/src/main/kotlin/base/boudicca/query/evaluator/util/EvaluatorUtil.kt
```diff
@@ -11,6 +11,8 @@ import java.time.temporal.ChronoUnit
 import java.util.concurrent.ConcurrentHashMap
 import kotlin.jvm.optionals.getOrNull
 
+private const val MILLIS_PER_HOUR = 1000.0 * 60.0 * 60.0
```
twatzl, line 14: Can't we do something liek Duration.hours(1).millis? :D

boudicca.events/eventcollectors/src/main/kotlin/events/boudicca/eventcollector/collectors/WissensturmCollector.kt
```diff
@@ -23,9 +23,14 @@ class WissensturmCollector : TwoStepEventCollector<Pair<String, Document>>("wiss
         var date = LocalDate.now(ZoneId.of("Europe/Vienna"))
 
         //only collect 6 months for now
-        for (i in 1..6) {
+        for (ignored in 1..6) {
```
twatzl, line 26: ignored?

boudicca.events/eventcollectors/src/main/kotlin/events/boudicca/eventcollector/collectors/LinzTermineCollector.kt
```diff
@@ -137,7 +137,7 @@ class LinzTermineCollector : EventCollector {
 
         var date = LocalDate.now(ZoneId.of("Europe/Vienna")).atStartOfDay()
         val links = mutableListOf<String>()
-        for (i in 1..(4 * 6)) {
+        for (ignored in 1..(4 * 6)) {
```
twatzl, line 140: ignored?

boudicca.base/common-model/src/main/kotlin/base/boudicca/model/structured/StructuredEntry.kt
kadhonn: there is no class here where we can take the name from :D Entries + StructuredEntries are only typealiases, so we need to set the name here manually

boudicca.base/common-model/src/main/kotlin/base/boudicca/model/structured/AbstractKey.kt
kadhonn: because there is a rule in detekt which says you should not do that and i agree :D

boudicca.base/fetcher-lib/src/main/kotlin/base/boudicca/fetcher/Fetcher.kt
kadhonn: not sure if that makes it better? i think it is ok this way, especially because all those parameters have default values anyway

boudicca.base/fetcher-lib/src/main/kotlin/base/boudicca/fetcher/Fetcher.kt
kadhonn: yes, at least for enums, pretty sure for constants as well

boudicca.base/publisher-event-html/src/main/kotlin/base/boudicca/publisher/event/html/service/EventService.kt
kadhonn: #683

boudicca.base/publisher-event-html/src/main/kotlin/base/boudicca/publisher/event/html/service/EventService.kt
kadhonn: how? i think it is long, but it is the same pattern every time, it does not make sense to split it

boudicca.base/query-lib/src/main/kotlin/base/boudicca/query/BoudiccaQueryBuilder.kt
kadhonn: it is not fixed. the method is called "equals" because the query is called like that. but normally "equals" methods have a different contract, you know like comparing objects and are not supposed to throw exceptions. this is a different kind of equals method so i suppress the rule

boudicca.base/query-lib/src/main/kotlin/base/boudicca/query/evaluator/OptimizingEvaluator.kt
kadhonn: i think it normally makes sense to have it, in most cases this is a good indicator that you should split the class

boudicca.base/query-lib/src/main/kotlin/base/boudicca/query/evaluator/SimpleEvaluator.kt
kadhonn: again, this is not a very complex method, it is just long because there is a long when in there. if you have any better suggestions tell me

boudicca.events/eventcollectors/src/main/kotlin/events/boudicca/eventcollector/collectors/WissensturmCollector.kt
kadhonn: i cannot use "_" as a name for a variable in kotlin in a for loop, so i have to use "ignored" to signify that i don't care about it. otherwise i have an unused variable

boudicca.events/eventcollectors/src/main/kotlin/events/boudicca/eventcollector/collectors/LinzTermineCollector.kt
kadhonn: i cannot use "_" as a name for a variable in kotlin in a for loop, so i have to use "ignored" to signify that i don't care about it. otherwise i have an unused variable

PR #682, opened 2025-03-21
WIP Abl/common model js

kadhonn:
alternative to tobis try to achieve a common model js + jvm build


boudicca.base/common-model/src/commonMain/kotlin/base/boudicca/format/URI.kt
```diff
@@ -0,0 +1,7 @@
+package base.boudicca.format
+
+
+expect class URI
+
+expect fun parseURI(uri: String): URI
```
twatzl, line 6: can we somehow group these parsing and formatting functions? right now they are floating around as global functions.

boudicca.base/common-model/src/jvmTest/kotlin/base/boudicca/keyfilters/KeySelectorTest.kt
```diff
@@ -18,8 +19,8 @@ class KeySelectorTest {
             .build()
             .selectSingle(testEntry())
 
-        assertTrue(result.isPresent)
-        assertEquals("description", result.get().first.toKeyString())
+        assertThat(result).isNotNull()
+        assertEquals("description", result!!.first.toKeyString())
```
twatzl, line 23: before we merge we should replace all those assertEquals as well with assertThat(...).isEqualTo() and then remove assertj and replace it entirely with assertk

boudicca.base/common-model/src/jvmTest/kotlin/base/boudicca/model/EventEntryConversionTest.kt
```diff
@@ -34,9 +37,10 @@ class EventEntryConversionTest {
             "description:format=markdown:lang=en" to "#my english markdown description"
         )
 
-        val event = Event.fromEntry(entry).get()
+        val event = Event.fromEntry(entry)
 
-        assertEquals(6, event.data.size)
+        assertThat(event).isNotNull()
+        assertEquals(6, event!!.data.size)
```
twatzl, line 43: before merging replace assertEquals

boudicca.base/query-lib/src/main/kotlin/base/boudicca/query/Utils.kt
```diff
@@ -45,10 +45,10 @@ object Utils {
                 )
             ).build()
         )
-        if (optionalDateText.isEmpty) {
+        if (optionalDateText == null) {
             return Instant.ofEpochMilli(0).atOffset(ZoneOffset.MIN)
         }
-        val dateText = optionalDateText.get().second
+        val dateText = optionalDateText.second
```
twatzl, line 51: we should rename optionalDateText

buildSrc/src/main/kotlin/boudicca-kotlin-multiplatform.gradle.kts
```diff
@@ -0,0 +1,62 @@
...
+import org.gradle.kotlin.dsl.withType
+
+/**
+ * plugin for applying the correct kotlin version
+ */
+
+plugins {
+    id("boudicca-base")
+    kotlin("multiplatform")
+    // disable until https://github.com/boudicca-events/boudicca.events/issues/663 is discussed
```
twatzl, line 12: will be fixed with rebase i guess?

gradle/libs.versions.toml
```diff
@@ -12,7 +12,7 @@ spring-boot = "3.4.3"
 swagger-core-version = "2.2.20"
 org-springframework-boot-gradle-plugin = "3.4.3"
 org-openapi-generator-gradle-plugin = "7.2.0"
-junit-jupiter = "5.12.1"
+junit-jupiter = "5.12.0"
```
twatzl, line 15: downgrade? why?

kotlin-js-store/yarn.lock
twatzl, line 1: should we use yarn.lock or package.json lock? i guess yarn is default for kmp?

boudicca.base/common-model/src/commonMain/kotlin/base/boudicca/format/URI.kt
kadhonn: i think we can group it as an object, i will try

boudicca.base/common-model/src/jvmTest/kotlin/base/boudicca/model/EventEntryConversionTest.kt
kadhonn: again, changing out assertj with assertk has nothing to do with this PR, so I will do it in another PR

gradle/libs.versions.toml
kadhonn: good question, probably some git weirdness, will change it back

kotlin-js-store/yarn.lock
kadhonn: i have no idea yet, i will look at the js setup/publishing now

boudicca.base/common-model/src/jvmTest/kotlin/base/boudicca/model/EventEntryConversionTest.kt
twatzl: but now we use assertk in oneline and just for one another line we have a different assertion. it would be good to stay consistent. it is not that many lines we would need to change. and it would keep up with the spirit of fixing things as we touch them.

twatzl:
I read more into the KMP stuff and it seems some adaptions with annotations are required.

I will put my activitiy on this ticket on hold for now as i have to prepare Museumsbahn-events for the upcoming seaseon

https://kotlinlang.org/docs/whatsnew20.html#improved-collection-interoperability
https://kotlinlang.org/docs/whatsnew2020.html#gradle
https://kotlinlang.org/docs/js-to-kotlin-interop.html

kadhonn:
i am not sure if we really should continue with this as it is. the JS code is suboptimal to say the least, also it really messes with the published packages.... it seems to me that the kotlin/js exporting is too beta, even for us.

the main issue is that kotlin/js is meant to be a project in itself, but not meant to make and publish a library

Issue #701, opened 2025-06-30
Frontend Analytics with some tool


Yolgie:
Look at:
* Matomo
* Ackee
* Plausible

compare, take a decision, add it to a separate docker compose

don't forget to add the info to the disclaimer/about us/impressum/privacy policy(todo) where necessary (check https://matomo.org/blog/2021/10/matomo-exempt-from-tracking-consent-in-france/)

PR #704, opened 2025-07-13
Abl/analytics

kadhonn:
added plausible and deployment to boudicca.events will be visible at https://plausible.monitoring.boudicca.events/boudicca.events


kadhonn:
would resolve #701

.gitignore
```diff
@@ -34,7 +34,6 @@ nb-configuration.xml
 *.rej
 
 # Local environment
-.env
```
twatzl, line 37: why? do we need some secrets? then it might make sense creating an additional secrets.env and add this one to gitignore

kadhonn: i did remove .env from gitignore, not add it
not sure why it was in there, i don't think we use it anywhere else

PR #796, opened 2025-10-04
add salonschiff fraeulein florentine collector

belehla:
closes #709

PR #798, opened 2025-10-05
add english months

kadhonn:


belehla:
I tested it with #796 (without am/pm time) and it worked perfectly, thanks! :)

PR #841, opened 2025-11-08
Fix kupfticket raw html in description & show line breaks in modal description

### LINKED ISSUE #840: Kupfticket events show raw HTML
I opened the event "TRANSILVANIA / CNTMPT / SERPERE / SEVEN CHURCHES" and the description starts with `<p><strong>`, which indicates HTML was not properly processed. I would expect to not see HTML tags in the description.

Image

belehla:
closes #840


belehla:
possible solution for rendering markdown in html --> #842

PR #842, opened 2025-11-09
ui: render markdown in modal & add kupfticket md description

belehla:
Example of parsed markdown: image

### DIFF
--- a/boudicca.base/publisher-event-html/src/main/resources/static/js/main.js
+++ b/boudicca.base/publisher-event-html/src/main/resources/static/js/main.js
@@ -19,6 +19,7 @@ document.addEventListener("DOMContentLoaded", () => {
   const map = document.getElementById("map");
   const multiselectFilterInputs = ["locationCities", "locationNames", "bandNames"];
   let lastFocusedEventCard = null;
+  const markdownForbiddenTags = ['img'];
 
 
   const openModal = (eventCard) => {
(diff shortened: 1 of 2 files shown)


boudicca.base/publisher-event-html/src/main/resources/templates/index.hbs
```diff
@@ -17,5 +17,6 @@
 
         <script src="js/main.js"></script>
         <script src="js/menu.js"></script>
+        <script src="https://cdn.jsdelivr.net/npm/marked/marked.min.js"></script>
```
kadhonn: please don't use CDNs but just add it into our static folder, it is not that big and we avoid external depenencies on our website, which can add instability + data leaks

boudicca.base/publisher-event-html/src/main/resources/static/js/main.js
```diff
@@ -395,4 +395,9 @@ document.addEventListener("DOMContentLoaded", () => {
   multiselectFilterInputs.forEach( (identifier) => {
     document.getElementById("filter-" + identifier).addEventListener("input", () => filterMultiselectFieldsByInput(identifier));
   })
+
+  document.querySelectorAll(".markdown").forEach(mdContainer => {
```
kadhonn: this only runs initially but not when you click the "load more" button

kadhonn: on second thought, we probably cannot use it as is anyway

PR #844, opened 2025-11-09
ui: add new search filters & show end date and address

belehla:
- new **filters** in drawer: tags, types and music genres
- show type, music genre and band names in modal
- add **color** to sport events
- show category colors in search filter
- show **date** without time if time is 00:00
- add end date in modal
- show **address** if no location name in card
- show address additionally in modal
- try to fix size of filters so that they move less around if things are selected


kadhonn (approved):
i love all of those changes! thanks for doing them <3

PR #860, opened 2025-12-07
chore(detekt): fix all wildcard import issues and add editorconfig to avoid them in the future

Yolgie:
currently fixing the detekt config, needs to be updated to 2.0 for our current kotlin version, in the process i'm fixing some detekt issues as separate branches where sensible :)

should be only reorganizing imports, formatting and some unused vars -> `_`


kadhonn (approved):

.editorconfig
```diff
@@ -3,3 +3,6 @@ root = true
 [*.{kt,kts}]
 insert_final_newline = true
 max_line_length = 120
+# Always use explicit imports (never use wildcard imports)
```
twatzl: why?

PR #861, opened 2025-12-07
upgrade detekt to v2.0.0-alpha.1, redo the detekt config, fix all open detekt issues

Yolgie:


boudicca.base/publisher-event-html/src/main/kotlin/base/boudicca/publisher/event/html/service/EventService.kt
```diff
@@ -36,8 +36,7 @@ import java.time.LocalDate
 import java.time.OffsetDateTime
 import java.time.ZoneId
 import java.time.format.DateTimeFormatter
-import java.util.Locale
-import java.util.Optional
+import java.util.*
```
kadhonn: star import?

.github/workflows/detekt.yml
```diff
@@ -10,19 +10,21 @@ on:
...
     steps:
-      - uses: actions/checkout@v6.0.1
-
-      - name: detekt
-        uses: alaegin/Detekt-Action@b43f1c93f8b0e1aedb1bf41719220eee3b2cd49b # v1.23.7
+      - name: Setup detekt
+        uses: peter-murray/setup-detekt@v3
         with:
-          github_token: ${{ secrets.github_token }}
+          detekt_version: v2.0.0-alpha.1
```
twatzl: boudicca is bleeding edge as always 😂

PR #890, opened 2025-12-12
Potential fix for code scanning alert no. 19: Untrusted Checkout TOCTOU

Yolgie:
Potential fix for [https://github.com/boudicca-events/boudicca.events/security/code-scanning/19](https://github.com/boudicca-events/boudicca.events/security/code-scanning/19)

To fix the problem, update the `actions/checkout` step so that the `ref` input uses the pull request's head SHA (the immutable commit hash) rather than the mutable branch name (`head.ref`). This ensures the workflow checks out exactly the code as it existed at the moment the comment was made, not any subsequently pushed code. You should modify the checkout step so that `ref` is set to `head.sha` from the PR API response (that's available via `fromJson(steps.pr.outputs.response).head.sha`). Additionally, update any step that refers to the branch name, where appropriate, to use the SHA for code checkout and build, but keep the branch name in commands or metadata as desired for deployment tracking. This change only needs to be made in the `.github/workflows/branchdeploy.yml` file.

---

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._

### DIFF
--- a/.github/workflows/branchdeploy.yml
+++ b/.github/workflows/branchdeploy.yml
@@ -33,7 +33,7 @@ jobs:
       - uses: actions/checkout@v6.0.1
         with:
           repository: ${{ fromJson(steps.pr.outputs.response).head.repo.full_name }}
-          ref: ${{ fromJson(steps.pr.outputs.response).head.ref }}
+          ref: ${{ fromJson(steps.pr.outputs.response).head.sha }}
       - name: Set up JDK 21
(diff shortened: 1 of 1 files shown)


kadhonn:
no this does not make sense. while it is true we have some sort of TOCTOU issue here, this suggested fix does nothing to prevent it. we would need to parse the comment history or something like that to make it really safe. or some github actions voodoo i am not familiar with

PR #994, opened 2026-05-31
add launch config for fetching online fetcher cache

kadhonn:


Yolgie:
i've fixed the added sonar issue and added a simple unit test

kadhonn:
now the linter fails... and we really need to configure the linter to print to the console, like this we cannot even see the error -.-

Yolgie:
should be fixed now, i would override the remaining sonarCloud (wants more code coverage) quality gate, everything sensible is tested now

Yolgie:
@kadhonn i think this is ready to merge, what do you say?

PR #997, opened 2026-05-31
event deduplication api definition

kadhonn:

PR #998, opened 2026-05-31
Abl/update clients

kadhonn:
updates to clients introducing interfaces to make mocks. depends on #997


Yolgie:
i'm rebasing this

PR #999, opened 2026-05-31
Enable gradle dependency locking

Yolgie:


buildSrc/src/main/kotlin/boudicca-base.gradle.kts
```diff
@@ -11,3 +11,13 @@ repositories {
     mavenCentral()
     mavenLocal()
 }
+
+dependencyLocking {
+    lockAllConfigurations()
+}
+
+tasks.register("writeLocks") {
```
kadhonn, line 19: why do you add this task? it seems unnecessary when we do not filter dependencies we want to lock. according to the doc we can simple call `gradle dependencies --write-locks` and it should update all locks: https://docs.gradle.org/current/userguide/dependency_locking.html#sec:lock-all-configurations-in-one-build-execution

kadhonn: this also makes it the same as in the buildSrc folder

buildSrc/src/main/kotlin/boudicca-base.gradle.kts
```diff
@@ -11,3 +11,13 @@ repositories {
     mavenCentral()
     mavenLocal()
 }
+
+dependencyLocking {
+    lockAllConfigurations()
```
kadhonn, line 16: we could enable strict locking: https://docs.gradle.org/current/userguide/dependency_locking.html#sec:fine-tuning-dependency-locking-behaviour-with-lock-mode

kadhonn (commented):
does our renovate bot still work with this change? if not i vote against it :(

as long as renovate still works this should be fairly noninvasive i think. although the benefit is also not quite high for us, since this feature is primarily designed to help with dynamic versions, which we do not use. so the only benefit would be if dependencies use dynamic versions, but i am not even sure if this is even possible once you publish an artifact...

PR #1060, opened 2026-07-19
render all properties

kadhonn:
i tried rendering all properties we have... but i am really not sure where to put them, if we should have them visible by default or if we should somehow format the keys better?
currently looks like this:
image


kadhonn:
!deploy

kadhonn:
!deploy

boudicca.base/publisher-event-html/src/main/resources/templates/layout/modal.hbs
```diff
@@ -1,3 +1,17 @@
 <div id="modal" class="modal">
-    <div id="modal-content"></div>
+    <div id="modal-container">
+        <span class="modal-buttons">
+            <button class="modal-toggle">
+                <svg height="32px" width="32px" viewBox="0 0 512 512" aria-label="Zeige alle Informationen">
+                    <use xlink:href="#menu"></use>
```
belehla: the menu button might be a little bit confusing here. maybe an info-i-icon or something like a document with a magnifying glass would be more suitable?

boudicca.base/publisher-event-html/src/main/resources/templates/events/event.hbs
```diff
@@ -57,173 +57,191 @@
...
-                            {{#if event.pictureCopyright}}
-                                <span class="image-copyright-info">&#169; {{event.pictureCopyright}}</span>
-                            {{/if}}
-                        {{else}}
-                            <svg title="Event Bild" viewBox="0 0 512 512" height="100px" aria-label="Event Bild">
-                                <use xlink:href="#{{#if event.category}}{{event.category}}{{else}}image{{/if}}"></use>
-                            </svg>
-                        {{/if}}
-                    </div>
+                    <h2 tabindex="0">{{event.name}}</h2>
```
belehla, line 63: on smaller screens, the new button almost overlaps with the heading here.. maybe add some margin on the right side of the heading?

Image

belehla (approved):

boudicca.base/publisher-event-html/src/main/resources/templates/layout/modal.hbs
kadhonn: yes you are right :D
i was just too lazy to figure out where/how to get new icons from <.<

boudicca.base/publisher-event-html/src/main/resources/templates/events/event.hbs
kadhonn: good catch!

kadhonn:
!deploy

